PR #4, opened 2023-05-11
Backward compatibility

nobu:
* [Use Array#& instead of Array#intersection](https://github.com/yui-knk/lrama/commit/2c8fd5dfe949462987069359d0f929d6e2a940bb)
* [Extract Lrama::Output#eval_template](https://github.com/yui-knk/lrama/commit/c681ad45ece215aa861e2dcafc569a8758f3ad42)
* [Support older ERb](https://github.com/yui-knk/lrama/commit/530294052642d78cfd1012de3d4c6ba82b15d6c8)
* [Remove indent at preprocessor directive for tool/ytab.sed](https://github.com/yui-knk/lrama/commit/8a621c8673393b065e34b55f48c83b5bb140d11a)

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: ['head', '3.2', '3.1', '3.0', '2.7']
+        ruby: ['head', '3.2', '3.1', '3.0', '2.7', '2.6', '2.5']
     steps:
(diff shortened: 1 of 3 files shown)


yui-knk:
Thank you. Just in case, which ruby version is the target of this PR, I will add CI and update `required_ruby_version` later.

nobu:
This is for 2.5.
As 2.4 does not support `Struct.new(keyword_init: true)`, it needs more works to support it.

PR #8, opened 2023-05-13
Add stdin mode

nobu:
When `-` is given as grammar file name, reads the grammar source from STDIN, and takes the next argument as the input file name.  This mode helps pre-process a grammar source.

### DIFF
--- a/lib/lrama/command.rb
+++ b/lib/lrama/command.rb
@@ -57,14 +57,18 @@ module Lrama
       end
 
       if !grammar_file
-        puts "File should be specified\n"
-        exit 1
+        abort "File should be specified\n"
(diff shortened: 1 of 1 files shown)


yui-knk:
Context: Current ruby/ruby build process is:

(1) line 927: Preprocess generates "parse.tmp.y" from "parse.c" by "tool/id2token.rb".
(2) line 928: Generate "y.tab.c" from "parse.tmp.y" by parser generator.
(3) line 930 & 932: Change file names in comments because temporal file name is used.

By this PR, we only need to pass preprocessed result to lrama command.

yui-knk (approved):

PR #13, opened 2023-05-16
Support `//` style comment

makenowjust:
I'm not familiar with YACC/Bison, so I'm not sure if this fix is the correct and expected one. I welcome any comments.


makenowjust:
Thank you, @yui-knk. I added a test-case for `//` style comment in `lex_common`.

yui-knk:
Thank you!

PR #14, opened 2023-05-16
Add a sample of calculator

mame:


sample/calc.y
```diff
@@ -0,0 +1,97 @@
...
+ * => 9
+ *
+ */
+
+%{
+#include <stdio.h>
+#include <stdlib.h>
+#include <ctype.h>
+
+#include "calc.h"
```
yui-knk, line 19: [note] Bison automatically insert `#include "calc.h"`, but lrama does not. I will revisit this point later.

yui-knk (approved):

PR #15, opened 2023-05-16
Add RBS and Steep check on CI

Little-Rubyist:
I introduced RBS (`bitmap.rb` only)

- Add RBS and Steep
- Run `steep check` with ruby HEAD on CI

I plan to eventually change `check "lib/lrama/bitmap.rb"` in steepfile to `lib`.


yui-knk:
Great! Thanks for your work 💯

PR #17, opened 2023-05-17
Added dependabot for GitHub Actions

ydah:
How about use dependabot like these?
- https://github.com/ruby/csv/pull/265
- https://github.com/ruby/fiddle/pull/116
- https://github.com/ruby/rexml/pull/89
- https://github.com/ruby/rss/pull/40
- https://github.com/ruby/xmlrpc/pull/31


yui-knk:
@ydah Thank you for your PR, LGTM. I will merge it.

PR #23, opened 2023-05-18
Add warning.rbs

akouryy:
Added warning.rbs ~~based on #19~~.

- The auto-generated type of `@out` was `IO`, but `@out` can be a String in [tests](https://github.com/ruby/lrama/blob/4b424415b49827758c99c8f4e74eb1dd71c80ef2/spec/lrama/context_spec.rb#L3).
- I could write the type of `message` to be `string & _ToS`, but I think it would be too complex.

### DIFF
--- a/Steepfile
+++ b/Steepfile
@@ -4,4 +4,5 @@ target :lib do
   signature "sig"
 
   check "lib/lrama/bitmap.rb"
+  check "lib/lrama/warning.rb"
 end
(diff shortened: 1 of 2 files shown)


Little-Rubyist:
Since `Warning` class has been removed from `typeprof-generated.rbs`, there seems no need to base #19 on it.
If we are not dependent on `gem_rbs_collection`, then we may check out from master.

akouryy:
Removed the dependency on #19.

yui-knk:
👍

PR #28, opened 2023-05-19
Refactoring lrama/states.rb

jinroq:
* Separated State class from lrama/states.rb file
* Separated Lrama::State::Reduce class and Lrama::State::Shift class into separate files


yui-knk (approved):

yui-knk:
Thank you!

PR #44, opened 2023-06-16
Merge error_recovery branch

alitaso345:
# Summary
- yui-knk-san has already implemented an error recovery mode, however, the 'error_recovery' branch has not yet been merged. 
- In this pull request, we will merge the 'error_recovery' branch, thereby enabling the retrieval of error recovery tokens.

# Changes Made
- The 'error_recovery' branch from the repository https://github.com/ruby/lrama/tree/error_recovery has been merged.
- template/bison/yacc.c file has been modified to ensure successful compilation
  - Added user arguments to functions and macros used by bison.


yui-knk:
Thanks for reviving the branch! I will merge this PR :)

These are future work we may need to tackle

1) Determine an interface for enable/disable error recovery
Error recovery is not zero cost. Therefore it might be disabled in many cases and enabled for some cases, e.g. when parser is used as backend of LSP. There are two options for this point.

1. Add lrama command line option or `%error-recovery` declaration.
2. Update `yyparse` or add callback function to determine error recovery enablement on runtime.

Option 1 is static then we need to build two parser in ruby/ruby. On the other hand option 2 is dynamic, so single parser can handle both use cases. I prefer option 2.

2) Determine an interface for max number of trial
At the moment `YYMAXREPAIR` macro controls the upper limit of insert operation so that the upper limit is determined on compile time. I think it's useful if the upper limit is configurable on runtime.

PR #47, opened 2023-06-23
Fix license notices

gemmaro:


yui-knk:
Thanks for the fix. I should correctly reflect the license written in https://github.com/ruby/lrama/blob/becdd6e4832d60b1033bd369248e9b14996a55e2/template/bison/yacc.c#L10-L13.

PR #51, opened 2023-07-08
Refactor lrama/state.rb 

alitaso345:
# Summary
- Separate  ResolvedConflict from lrama/state.rb

# Changes Made
- Add lrama/state/resolved_conflict.rb

PR #50, opened 2023-07-06
Refactor states.rb and grammar.rb

alitaso345:
# Summary
- I have refactored `states.rb` and `grammar.rb`.
- Groups of Structs with methods have been separated into different files as Classes inheriting from Struct.

# Changes Made
- `states.rb`
    - The `Item` has been separated into a different file.
- `grammar.rb`
    - The following Structs have been separated into different files:
    - `Code`, `ErrorToken`, `Precedence`, `Printer`, `Reference`, `Rule`, `Symbol`, `Union`

PR #55, opened 2023-07-27
Refactor report.rb to use one module per source file

ydah:
This PR refactor report.rb to use one module per source file.

As with the following PRs, this PR change to one file per module.
- https://github.com/ruby/lrama/pull/50
- https://github.com/ruby/lrama/pull/51


yui-knk:
Thank you!

PR #60, opened 2023-07-31
Raise proper error at reference with no declared type

nobu:
When using reference with no declared type mistakenly, it raise `NoMethodError` with ``undefined method `s_value' for nil`` currently.
Of course it is wrong code, but the message does not seem intentional.

```
./lib/lrama/grammar/code.rb:77:in `block in translated_user_code'
./lib/lrama/grammar/code.rb:63:in `each'
./lib/lrama/grammar/code.rb:63:in `translated_user_code'
./lib/lrama/grammar/code.rb:14:in `translated_code'
./lib/lrama/grammar/rule.rb:30:in `translated_code'
```

### DIFF
--- a/lib/lrama/grammar/code.rb
+++ b/lib/lrama/grammar/code.rb
@@ -8,10 +8,10 @@ module Lrama
       def_delegators "token_code", :s_value, :line, :column, :references
 
       # $$, $n, @$, @n is translated to C code
-      def translated_code
+      def translated_code(rule = nil)
         case type
         when :user_code
-          translated_user_code
+          translated_user_code(rule)
         when :initial_action
           translated_initial_action_code
         end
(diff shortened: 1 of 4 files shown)


nobu:
I'm not sure that this is a right fix; passing rule's name to `Lrama::Grammar::Code#translated_user_code` feels invasive.

yui-knk:
I created https://github.com/ruby/lrama/pull/61. I hope https://github.com/ruby/lrama/pull/61/files#diff-f8748936dd72d99dbcf38b4e4c65325beb000c42d2c1779dd3ca1bcb57bdd0deR1375 covers all cases. Could you check it when you have time?

nobu:
Close in favor of #61.

PR #61, opened 2023-07-31
Raise error when referred nterm, term and action type is not declared

yui-knk:


nobu (approved):

yui-knk:
Thanks for the confirmation.

PR #62, opened 2023-07-31
Replace parser with one made by Racc

junk0612:
Generate the parser of grammar file with Racc and replace it from original parser.

### TODOs

- [x] Enable to parse the Ruby's `parse.y`
- [x] Assemble the same AST as the original parser
- [x] ~~Replace the parser with the original~~
    - Move to future work.


lib/lrama/grammar.rb
```diff
@@ -306,6 +306,48 @@ def nterms
...
+        rhs.each_with_index do |token, index|
+          next if token.type != Lrama::Lexer::Token::User_code
+
+          scanner = StringScanner.new(token.s_value)
+          references = []
+
+          while !scanner.eos? do
+            start = scanner.pos
+            case
+            when scanner.scan(/\$(<[a-zA-Z0-9_]+>)?\$/) # $$, $<long>$
```
krzysiek1507, line 320: If don't need to use the result of `scan`, you can try to use `skip` and save a lot of allocations and improve the performance

krzysiek1507: Similar what I did in this PR https://github.com/ruby/lrama/pull/102

yui-knk:
:shipit:

yui-knk:
I'm excited to merge this PR!
Hand written parser is replaced by LR parser generated by Racc.
Now, Ruby runs Racc, Racc generates LR parser for Lrama and Lrama generates LR parser for Ruby!!!

Greatly appreciate your hard work 🎉

PR #64, opened 2023-08-01
Refactor Lrama::Grammar::Symbol

Maumagnaguagno:
Simplify display_name and enum_name methods by removing if-else and using non-word character class.


yui-knk:
Thanks!

PR #65, opened 2023-08-01
Refactor Lrama::Grammar::Rule

Maumagnaguagno:
Prefer safe navigation in precedence and translated_code, and symbol to proc in as_comment.


yui-knk:
Thanks!

PR #66, opened 2023-08-04
Prefer transform_values in Lrama::States

Maumagnaguagno:
Replace each with transform_values in direct_read_sets, read_sets, follow_sets and la methods.


yui-knk:
Thank you!

PR #67, opened 2023-08-04
Moved Grammar unit tests

lucasmenezesds:
### Main Changes

- [Commit](https://github.com/ruby/lrama/commit/22bb1c46ccfa7f612c8ac3a6c4f4708af0f75db0)
  - Moved Grammar unit tests for it's own folder and changed the name to represent what was really testing (symbol_spec)
  - Added the setting for randomizing the orders the unit tests run
  - Added the setting to allow focusing tests (but there is an attention point)
  
 - [Commit](https://github.com/ruby/lrama/commit/4a199e539efb6ccb7ef9120c564b4352505180ec)
   - Removed the constant definition to the let statement instead
   - Removed extra describe (`RSpec.describe Lrama::Grammar do`)
   - Changed to use `described_class` instead of `Lrama::Grammar::Symbol`
   
- [Commit](https://github.com/ruby/lrama/commit/f668c48ad79421f1182019302bf515567c6a701c)
  - Created a simple unit test for Code#translated_code


spec/spec_helper.rb
```diff
@@ -25,6 +25,13 @@
   config.expect_with :rspec do |c|
     c.syntax = :expect
   end
+
+  # Randomize specs run order.
+  config.order = :random
+  Kernel.srand config.seed
+
+  # Allow to limit the run of the specs
+  config.filter_run_when_matching :focus
```
lucasmenezesds, line 34: The attention point for having this enabled is that if someone forgets the focus on the code, then this would be a problem since most of the tests would be skipped.

So I'd like your opinion on this matter

spec/lrama/grammar/code_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+      it "calls #translated_initial_action_code" do
+        code = described_class.new(type: :initial_action, token_code: initial_act_token)
+
+        expect(code).to receive(:translated_initial_action_code)
+
+        code.translated_code
+      end
+    end
+  end
+end
```
yui-knk, line 27: Please add new line on the end of a file.

spec/spec_helper.rb
yui-knk: I prefer to NOT set this configuration because it's risky and costly to check each PR does not introduce focus method (e.g. `fit`). Could you remove this configuration?

spec/spec_helper.rb
```diff
@@ -25,6 +25,13 @@
   config.expect_with :rspec do |c|
     c.syntax = :expect
   end
+
+  # Randomize specs run order.
+  config.order = :random
+  Kernel.srand config.seed
```
yui-knk, line 31: I feel this code bit tricky. I guess the intention is to fix seed of Kernel#rand. However IIRC at the moment there is no dependency for such randomness.

yui-knk:
Thank you!!

PR #77, opened 2023-08-08
Early break and safe navigation to Lrama::Digraph#traverse

Maumagnaguagno:
Skip ``@result`` reassignment to the same value.


yui-knk:
Thank you!

PR #79, opened 2023-08-09
Extract Struct into another file 

ydah:
Follow up: https://github.com/ruby/lrama/pull/78


lib/lrama/lexer/token.rb
```diff
@@ -1,7 +1,7 @@
 module Lrama
   class Lexer
     class Token < Struct.new(:type, :s_value, :alias, keyword_init: true)
-      Type = Struct.new(:id, :name, keyword_init: true)
+      require 'lrama/lexer/token/type'
```
yui-knk, line 4: Could you move this "require" to the first line of this file?

ydah: Done. I updated this PR.

yui-knk:
Thank you!

PR #88, opened 2023-08-22
Fix lexing `@10` in user codes correctly

makenowjust:
A regexp `/@(\d)+/` seems wrong because this only captures the last digit in `$1`. It may not be intended.

This fixes the regexp and also adds new test cases for some tokens in user codes.


yui-knk:
Thank you!!

PR #97, opened 2023-09-12
Improve output when executing the help command

ydah:
This PR improve output when executing the help command.
In addition, add `Options` and `OptionParser` classes for refactoring.

## Before
```
❯ bundle exec lrama --help               
Usage: lrama [options]
    -V, --version
    -S, --skeleton=FILE
    -t
    -h, --header=[FILE]
    -d
    -r, --report=THINGS
        --report-file=FILE
    -v
    -o, --output=FILE
        --trace=THINGS
    -e
```

## After
```
❯ bundle exec lrama --help
Lrama is LALR (1) parser generator written by Ruby.

Usage: lrama [options] FILE

Tuning the Parser:
    -S, --skeleton=FILE              specify the skeleton to use
    -t                               reserved, do nothing

Output:
    -h, --header=[FILE]              also produce a header file named FILE
    -d                               also produce a header file
    -r, --report=THINGS              also produce details on the automaton
        --report-file=FILE           also produce details on the automaton output to a file named FILE
    -o, --output=FILE                leave output to FILE
        --trace=THINGS               also output trace logs at runtime
    -v                               reserved, do nothing

Error Recovery:
    -e                               enable error recovery

Other options:
    -V, --version                    output version information and exit
        --help                       display this help and exit
```


lib/lrama/option_parser.rb
```diff
@@ -0,0 +1,124 @@
...
+    end
+
+    private
+
+    def parse_by_option_parser(argv)
+      ::OptionParser.new do |o|
+        o.banner = <<~BANNER
+          Lrama is LALR (1) parser generator written by Ruby.
+
+          Usage: #{$0} [options] FILE
```
yui-knk, line 52: `$0` sometimes verbosely print full path of program file like your example in PR description. I think it's better to print only base name ("lrama").

In this case, the name ("lrama") is static then it's ok to rewrite this banner like

```ruby
Usage: lrama [options] FILE
```

Default value of `program_name` is `File.basename($0, '.*')`. https://github.com/ruby/optparse/blob/a291ef5c26445d423bf0223b0cb64340df18f0cb/lib/optparse.rb#L1248

Or another approach is using `#banner=` and `#banner` like

```ruby
      opt = OptionParser.new do |parser|
        parser.banner += " FILE"
      end
```

https://github.com/ruby/optparse/blob/a291ef5c26445d423bf0223b0cb64340df18f0cb/lib/optparse.rb#L104 is an example.

ydah: Indeed it is. I will change it to use the static "lrama" string.

yui-knk:
Thanks for the great improvement!

PR #99, opened 2023-09-14
Add digraph.rbs

Little-Rubyist:
Added type information to digraph.rb.

You may need to modify the actual code in some cases, since nil is entered in some places.
Currently nil is allowed for type checking.


yui-knk:
> Currently nil is allowed for type checking.

Sounds reasonable. Thank you!

PR #100, opened 2023-09-14
Make sure `z` to be not nil

yui-knk:
`Digraph#traverse` push argment first. This means while loop stops before stack to be empty because it will be break if argment is popped. Checking the result of `@stack.pop` in while condition expr, so that we can ensure `z` to be not nil on semantics level.


yui-knk:
This is a follow up of https://github.com/ruby/lrama/pull/99.

PR #101, opened 2023-09-14
Iterate over symbols to get the number to avoid array creation

krzysiek1507:
A tiny optimisation that reduces the memory usage by eliminating an array that it not used later anyway.

For parsing `spec/fixtures/context/basic.y` it reduces one object:

Before:
```
Calculating -------------------------------------
              master   133.668k memsize (     2.481k retained)
                         1.976k objects (     1.000  retained)
                        50.000  strings (     1.000  retained)
```

After:
```
Calculating -------------------------------------
              master   133.348k memsize (     2.481k retained)
                         1.975k objects (     1.000  retained)
                        50.000  strings (     1.000  retained)
```

### DIFF
--- a/lib/lrama/grammar.rb
+++ b/lib/lrama/grammar.rb
@@ -663,8 +663,8 @@ module Lrama
       nterm_token_id = 0
       used_numbers = {}
 
-      @symbols.map(&:number).each do |n|
-        used_numbers[n] = true
+      @symbols.each do |symbol|
+        used_numbers[symbol.number] = true
       end
 
       (@symbols.select(&:term?) + @symbols.select(&:nterm?)).each do |sym|


krzysiek1507:
Hey @yui-knk is this PR worth merging?

yui-knk:
@krzysiek1507 Thanks for your PR. However I think this is not bottleneck of current Lrama. If it's not bottleneck, I want to keep current codes because it's intuitive for me.

By the way, this is how to take profiling for real world use case, generating parser for ruby/ruby.

### 1. Create `parse.tmp.y` in ruby/ruby

```shell
$ ruby tool/id2token.rb parse.y > parse.tmp.y
$ cp parse.tmp.y dir/lrama/tmp
```

### 2. Enable Profiler

```diff
diff --git a/exe/lrama b/exe/lrama
index ba5fb06..2497178 100755
--- a/exe/lrama
+++ b/exe/lrama
@@ -3,4 +3,6 @@
 $LOAD_PATH << File.join(__dir__, "../lib")
 require "lrama"

-Lrama::Command.new.run(ARGV.dup)
+Lrama::Report::Profile.report_profile do
+  Lrama::Command.new.run(ARGV.dup)
+end
```

### 3. Run Lrama

```shell
$ bundle exec exe/lrama -d -o parse.tmp.c -h tmp/parse.tmp.y
```

### 4. Generate Flamegraph

```shell
$ stackprof --d3-flamegraph tmp/stackprof-cpu-myapp.dump > tmp/flamegraph.html
```

PR #102, opened 2023-09-14
Use skip instead of scan in Lexer

krzysiek1507:
It looks like we can use `skip` instead of `scan` in Lexer and save quite a lot allocations. We don't use the result of `scan` so we don't need the allocations.

For parsing `spec/fixtures/context/basic.y`:

Before:
```
Warming up --------------------------------------
              master    59.000  i/100ms
Calculating -------------------------------------
              master    606.783  (± 2.8%) i/s -      3.068k in   5.060887s
Calculating -------------------------------------
              master   131.059k memsize (     0.000  retained)
                         1.975k objects (     0.000  retained)
                        50.000  strings (     0.000  retained)
```

After:
```
Warming up --------------------------------------
              master    61.000  i/100ms
Calculating -------------------------------------
              master    615.521  (± 1.3%) i/s -      3.111k in   5.055099s
Calculating -------------------------------------
              master   119.339k memsize (     0.000  retained)
                         1.682k objects (     0.000  retained)
                        50.000  strings (     0.000  retained)
```


krzysiek1507:
Hey @yui-knk is this PR worth merging? I have a few more performance fixes in my local repo

PR #104, opened 2023-10-05
Change to automatically insert `#include` of header files when they are specified

ydah:
Resolve: https://github.com/ruby/lrama/pull/14#discussion_r1195204801

This PR change to automatically insert `#include` of header files when they are specified.
Bison will automatically embed `#include` header files when run with the `-H, --header=[FILE]` or `-d` options.
So, we make header file is include automatically when `-h`, `--header` or `-d` option is passed.

~We found that if Lrama was made to behave the same as Bison as is, it would [fail to build ruby](https://github.com/ruby/lrama/actions/runs/6520458872/job/17708035260).
The use of the `-H, --header=[FILE]` or `-d` options to include header files seemed to cause problems in cases like ripper.y, where ripper.h was not included from ripper.c.
So we added the `--auto-include` option to automatically add an include statement to the header file when this option is specified. How about it?~

### DIFF
--- a/template/bison/yacc.c
+++ b/template/bison/yacc.c
@@ -101,8 +101,12 @@
 # endif
 
 <%# b4_header_include_if -%>
+<%- if output.include_header -%>
+#include "<%= output.include_header %>"
+<%- else -%>
 /* Use api.header.include to #include this header
    instead of duplicating it here.  */
+<%- end -%>
 <%# b4_shared_declarations -%>
   <%-# b4_cpp_guard_open([b4_spec_mapped_header_file]) -%>
     <%- if output.spec_mapped_header_file -%>
--- a/README.md
+++ b/README.md
@@ -36,13 +36,8 @@ $ lrama -d sample/parse.y
 ```
 
 ```shell
-# "y.tab.c" and "y.tab.h" are generated and added `#include "y.tab.h"`
-$ lrama -d sample/parse.y --auto-include
-```
-
-```shell
-# "calc", "calc.c", and "calc.h" are generated and added `#include "calc.h"`
-$ lrama -d sample/calc.y -o calc.c --auto-include && gcc -Wall calc.c -o calc && ./calc
+# "calc", "calc.c", and "calc.h" are generated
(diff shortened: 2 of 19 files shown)


yui-knk:
Thanks for the PR. My expectation is that a header file is included automatically when `-h`, `--header` or `-d` option is passed.
I suspect the cause of ripper build failure is that `#include` is placed on very beginning of output file so that a lot of declarations, e.g. `VALUE`, `NODE` and so on, are not defined when the header file is included.
Could you try the below?

1. Rebase current master to reflect https://github.com/ruby/lrama/pull/133
2. Update https://github.com/ruby/lrama/blob/21e405abe539017f6d12243c4cd1f4a786fe4754/template/bison/yacc.c#L104 like below?

```
<%# b4_header_include_if -%>
<%- if output.header_file_path -%>
#include <%= option.include_header %>
<%- else -%>
/* Use api.header.include to #include this header
   instead of duplicating it here.  */
...
  <%-# b4_percent_code_get([[provides]]). %code is not supported -%>
  <%-# b4_cpp_guard_close([b4_spec_mapped_header_file]) -%>
    <%- if output.spec_mapped_header_file -%>
#endif /* !<%= output.b4_cpp_guard__b4_spec_mapped_header_file %>  */
    <%- end -%>
<%- end -%>
<%# b4_declare_symbol_enum -%>
```

ydah:
Thank you so much! I updated this PR.

yui-knk:
:shipit:

template/bison/yacc.c
```diff
@@ -101,8 +101,12 @@
 # endif
 
 <%# b4_header_include_if -%>
+<%- if output.include_header -%>
+#include "<%= output.include_header %>"
+<%- else -%>
 /* Use api.header.include to #include this header
    instead of duplicating it here.  */
+<%- end -%>
```
yui-knk, line 109: Oh, I overlooked the location of `<%- end -%>` is wrong...
I will fix it within https://github.com/ruby/lrama/pull/135.

ydah: Oh, my apologies. Thank you for fixing it 🙇🏻

PR #114, opened 2023-10-12
add lexer/token/type.rbs

Little-Rubyist:
I was planning to type-check `symbol.rb`, but I realized that I need `grammar.rb`, `lexer/token.rb`, and `lexer/token/type.rb`.
I'm working on preparing them in order, and this is going to be the first step in that process.

### DIFF
--- a/sig/lrama/lexer/token/type.rbs
+++ b/sig/lrama/lexer/token/type.rbs
@@ -1,0 +1,15 @@
+module Lrama
+  class Lexer
+    class Token
+      attr_accessor type: Type
+      attr_accessor s_value: String
+      attr_accessor alias: String
+      attr_accessor keyword_init: bool
+      class Type
+        attr_accessor id: Integer
+        attr_accessor name: String
+        attr_accessor keyword_init: bool
+      end
+    end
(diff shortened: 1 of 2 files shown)


yui-knk:
LGTM! Thank you!

sig/lrama/lexer/token/type.rbs
```diff
@@ -0,0 +1,15 @@
+module Lrama
+  class Lexer
+    class Token
+      attr_accessor type: Type
+      attr_accessor s_value: String
+      attr_accessor alias: String
+      attr_accessor keyword_init: bool
```
yui-knk, line 7: `keyword_init` of `Struct.new` doesn't define `keyword_init` accessor on the class it creates, I noticed it after merge.

```ruby
>> s = Struce.new(:a, keyword_init: true)
>> s.new(a: 1).a
=> 1
>> s.new(a: 1).keyword_init
(irb):6:in `<main>': undefined method `keyword_init' for an instance of #<Class:0x0000000105b9df38>(keyword_init: true) (NoMethodError)
```

PR #116, opened 2023-10-13
fix lexer/token/type.rbs

Little-Rubyist:
I've made the requested revisions to the previously PR(#114).
Also, I received guidance on #types in ruby-jp and added type definitions for the initialize method.


yui-knk:
Go ahead! Thank you!

PR #118, opened 2023-10-14
Build Lrama::Lexer::Token in Lrama::NewLexer#next_token

junk0612:
Token generation was moved from Parser to Lexer in order to simplify the description of Parser.


yui-knk:
LGTM, thank you!

PR #123, opened 2023-10-15
Fix a freeze when specified `grammar_file` is '-' with file path

ydah:


ydah:
Sorry. I was mistaken.

PR #127, opened 2023-10-16
Use NewParser in spec files if possible

yui-knk:
We are migrating Lrama's parser from hand written parser to LR parser generated by Racc. Replacing Parser with NewParser to prevent regression.


yui-knk:
@junk0612 What do you think about this change?

junk0612:
Thank you for replacing.

PR #137, opened 2023-10-18
Replace NewParser with Parser #6

junk0612:
- Parse some escape characters


yui-knk:
There is no skipped test case in "new_parser_spec.rb"  💯

PR #139, opened 2023-10-19
Allow `-H` option to also generate a header file with the specified name

ydah:
In bison it is `-H` but in lrama it is `-h`.
How about making it available in `-H` as well?

Also, how about indicating that `-h` is deprecated in the help command as a soft deprecation?


yui-knk:
Sounds reasonable.

PR #141, opened 2023-10-19
Add a newline at the end of C_DECLARATION in epilogue

ydah:
Resolve: https://github.com/yui-knk/ruby/actions/runs/6545361161/job/17773670072


yui-knk:
I think there is 3 approach for this problem.

1. Add a new line to template/bison/yacc.c (this PR)
2. Use `\z` for `end_symbol` of `epilogue_opt`
3. Add another lexer state for epilogue because epilogue can be simply treated line by line

This is most simple approach, then I will try this first.

PR #142, opened 2023-10-19
Replace NewParser with Parser #Fin

junk0612:
- Remove old hand-made parser
- Rename `Lrama::NewParser` to `Lrama::Parser`


yui-knk:
Could you also update this line https://github.com/ruby/lrama/blob/0be2334ffb6e0d27a91c760bffa5de8f7e6823a7/.github/workflows/test.yaml#L61 ?

yui-knk:
💯

PR #149, opened 2023-10-22
Replace NewLexer with Lexer

junk0612:
- Add test cases to NewLexer
- Remove old Lexer class and replace NewLexer with it


yui-knk:
:shipit:

PR #152, opened 2023-10-23
Add test case for `#translated_printer_code`

ydah:


spec/lrama/grammar/code_spec.rb
```diff
@@ -24,4 +24,63 @@
       end
     end
   end
+
+  describe "#translated_printer_code" do
+    let(:printer_token) { token_class.new(type: token_class::P_printer, s_value: '<val>') }
```
yui-knk, line 29: Could you change `token_class::P_printer` to `token_class::User_code`?
Currently token of code passed to printer is `Token::User_code`.

* The second argument of `build_code` is `val[3]`. https://github.com/ruby/lrama/blob/d9418cde04a93ac628d45e4b482054a345accbcf/parser.y#L101
* `val[3]` is `C_DECLARATION` (0-origin). https://github.com/ruby/lrama/blob/d9418cde04a93ac628d45e4b482054a345accbcf/parser.y#L94
* Token type of `C_DECLARATION` is `Token::User_code` https://github.com/ruby/lrama/blob/d9418cde04a93ac628d45e4b482054a345accbcf/lib/lrama/lexer.rb#L118-L124

ydah: Indeed it is. Thank you! I updated this PR.

yui-knk:
Thank you!

PR #161, opened 2023-10-27
add lexer/token.rbs

Little-Rubyist:
relation: #114

### DIFF
--- a/sig/lrama/lexer/token.rbs
+++ b/sig/lrama/lexer/token.rbs
@@ -1,0 +1,47 @@
+module Lrama
+  class Lexer
+    class Token
+
+      attr_accessor line: Integer
+      attr_accessor column: Integer
+      attr_accessor referred: bool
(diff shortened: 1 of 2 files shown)


sig/lrama/lexer/token.rbs
```diff
@@ -0,0 +1,47 @@
...
+    class Token
+
+      attr_accessor line: Integer
+      attr_accessor column: Integer
+      attr_accessor referred: bool
+      attr_accessor references: Array[[Symbol, (String|Integer), Token?, Integer, Integer]]
+
+      def to_s: () -> String
+      def referred_by?: (String string) -> bool
+      def ==: (Token other) -> bool
```
Little-Rubyist, line 12: I'm only worried about the argument type of the `==` method.

yui-knk: The code can handle other type objects however I expect `other` is Token. Therefore let's try `Token` for the argument type.

yui-knk:
LGTM, thank you!

PR #170, opened 2023-10-30
Extract C_DECLARATION end process to a private method

ydah:
Follow up: https://github.com/ruby/lrama/pull/168/commits/9eff5048c160dc5b339bdfbdbf6e98fe0aff421f

### DIFF
--- a/lib/lrama/parser.rb
+++ b/lib/lrama/parser.rb
@@ -719,7 +719,7 @@ def begin_curly_brace
   @lexer.end_symbol = '}'
 end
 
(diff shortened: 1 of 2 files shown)


yui-knk:
These are clean up processes for `C_DECLARATION` because actions (`{...}`) after `C_DECLARATION` are executed after  `C_DECLARATION` is determined. Could you rethink the name of the method?

ydah:
Oops, sorry. I change to use `end`. What do you think?

PR #185, opened 2023-11-04
Add the starting position to the location information

ydah:
```
❯ exe/lrama -d test.y                         
parser.y:400:in `on_error': a.y:5:7: parse error on value #<struct Lrama::Lexer::Token::Ident s_value="invalid", alias_name=nil> (IDENTIFIER) (Racc::ParseError)
%expect invalid
        ^^^^^^^
        from racc/parser.rb:276:in `_racc_do_parse_c'
        from racc/parser.rb:276:in `do_parse'
        from parser.y:386:in `block in parse'
        from /ydah/lrama/lib/lrama/report/duration.rb:14:in `report_duration'
        from parser.y:381:in `parse'
        from /ydah/lrama/lib/lrama/command.rb:11:in `run'
        from exe/lrama:6:in `<main>'
```

### DIFF
--- a/lib/lrama/parser.rb
+++ b/lib/lrama/parser.rb
@@ -688,11 +688,10 @@ def next_token
 end
 
 def on_error(error_token_id, error_value, value_stack)
-  source = @text.split("\n")[error_value.line - 1]
   raise ParseError, <<~ERROR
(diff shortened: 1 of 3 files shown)


yui-knk:
Thank you!

PR #212, opened 2023-11-09
Support for whitespace before parentheses in parameterizing rules

ydah:


yui-knk:
Thanks for the PR, however I think current implementation is a little complex.
My proposal for improvement is

## `parser.y`

Change these parameterizing rules

* `rhs: rhs parameterizing_suffix` to `rhs: rhs IDENTIFIER parameterizing_suffix`
* `rhs: parameterizing_prefix "(" rhs ")"` to `rhs: rhs IDENTIFIER "(" symbol ")"`
* `rhs: parameterizing_separated_prefix "(" symbol "," rhs ")"` to `rhs: rhs IDENTIFIER "(" symbol "," symbol ")"`

Where `IDENTIFIER` means parameterizing rule name like `nonempty_list` and `symbol` means `IDENTIFIER` (`number`), `CHARACTER` (`','`) and `string_as_id` (`"="`).

## `Parameterizing`

Add `attr_accessor :args` (Array) to `Lrama::Lexer::Token::Parameterizing` so that it can hold arbitrary number of arguments .

## `RuleBuilder`

Update `RuleBuilder#expand_parameterizing_rules` to check `Parameterizing#s_value` (rule name) and decide which expand logic.

ydah:
Thank you for your proposal. It certainly makes sense, so I have updated this PR. What do you think?

lib/lrama/lexer.rb
```diff
@@ -29,23 +29,19 @@ class Lexer
...
-      'separated_list\(',
-    ]
 
     def initialize(text)
       @scanner = StringScanner.new(text)
       @head_column = @head = @scanner.pos
       @head_line = @line = 1
       @status = :initial
       @end_symbol = nil
+      @lex_stack = []
```
yui-knk, line 39: No value is pushed to this instance variable, could you remove this line and the first line of `#next_token`?

yui-knk:
Left a minor comment, other seem good to me.

lib/lrama/lexer.rb
ydah: Oops sorry, I deleted these.

yui-knk:
Thanks!

PR #218, opened 2023-11-12
Add `spec` and `steep check` to Rake's default tasks

ydah:
This PR is add `spec` and `steep check` to Rake's default tasks.
How about just running the following command to check them all together?

```sh
❯ bundle exec rake   
```


yui-knk:
LGTM!

PR #237, opened 2023-11-14
Change regexp to string for more readable

ydah:


ydah:
Well, in Ruby 2.5, a strong argument type String (expected Regexp) occurs, so the replacement is not possible. I'll close it.

PR #239, opened 2023-11-15
Add rbs for token

ydah:


sig/lrama/lexer/token/user_code.rbs
```diff
@@ -2,6 +2,8 @@ module Lrama
   class Lexer
     class Token
       class UserCode < Token
+        self.@references: Array[Lrama::Grammar::Reference]
```
yui-knk, line 5: This is an instance variable then `@references` is correct.

https://github.com/ruby/rbs/blob/47a57ff61e619dad6a34307a63c4f926cbf46e5e/docs/syntax.md#ivar-definition

ydah: Oops, sorry. I updated this PR.

yui-knk:
👍

PR #240, opened 2023-11-16
Organize rbs for `Lrama::Grammar::RuleBuilder`

ydah:


sig/lrama/grammar/rule_builder.rbs
```diff
@@ -7,26 +7,35 @@ module Lrama
       attr_reader user_code: Lexer::Token::UserCode?
       attr_reader precedence_sym: Lexer::Token?
 
+      @rule_counter: Counter
+      @midrule_action_counter: Counter
+      @position_in_original_rule_rhs: Integer?
+      @skip_preprocess_references: bool?
```
yui-knk, line 13: This is not optional, then `bool` (without `?`) is correct.

yui-knk:
Left a 1 comment, other seems good.

sig/lrama/grammar/rule_builder.rbs
ydah: Indeed it is. I updated this PR/

yui-knk:
👍

PR #242, opened 2023-11-16
Change lexer regex to be simpler

ydah:


lib/lrama/lexer.rb
```diff
@@ -71,8 +71,7 @@ def lex_token
           # noop
         when @scanner.scan(/\/\*/)
           lex_comment
-        when @scanner.scan(/\/\//)
-          @scanner.scan_until(/\n/)
+        when @scanner.scan(/\/\/.*?\n/)
```
yui-knk, line 74: This is changed by https://github.com/ruby/lrama/pull/245. Could you rebase (or merge) master branch and use current master implementation?

ydah: Sorry for the delay. I updated this PR.

yui-knk:
Thank you!

lib/lrama/lexer.rb
yui-knk: No worries!

PR #245, opened 2023-11-17
Lex a line comment without newline correctly

makenowjust:
Currently, the lexer produces an identifier token for `// foo <EOF>` because it only skips `//` when a line comment has no newline.

We can create such a funny example:

```
$ ruby -e 'print ?/*20_000' > p.y
$ exe/lrama p.y
parser.y:459:in `on_error': p.y:10001:-1: parse error on value "$" ($end) (Racc::ParseError)

...
```

Why the reported line number is `10001`??


yui-knk:
Seems good, thank you!

PR #260, opened 2023-11-22
Add steepcheck for lib/lrama/grammar/parameterizing_rules/**/*.rb

ydah:
This PR add steepcheck for lib/lrama/grammar/parameterizing_rules/**/*.rb


sig/lrama/grammar/parameterizing_rules/builder/list.rbs
```diff
@@ -0,0 +1,11 @@
+module Lrama
+  class Grammar
+    class ParameterizingRules
+      class Builder
+        class List < Base
+          def build: -> Array[Rule]
```
yui-knk, line 6: I think these sub classes inherit type definition of `#build` from `Base` class. Is this, and following type definition of `#build`, needed?

Steepfile
```diff
@@ -8,6 +8,13 @@ target :lib do
   check "lib/lrama/grammar/code.rb"
   check "lib/lrama/grammar/counter.rb"
   check "lib/lrama/grammar/error_token.rb"
+  check "lib/lrama/grammar/parameterizing_rules/builder.rb"
```
Little-Rubyist, line 11: It would be better to write `"lib/lrama/grammar/parameterizing_rules"` to detect when files are added to the directory.

yui-knk: Sound reasonable.

ydah: Indeed it is. Thank you! I updated it.

sig/lrama/grammar/parameterizing_rules/builder/list.rbs
ydah: Indeed it is. Thank you! I deleted it.

yui-knk:
Thank you!

PR #272, opened 2023-11-28
Add support for type specification in parameterizing rules

ydah:
This PR add support for type specification in parameterizing rules.


yui-knk:
👍 +1

PR #273, opened 2023-11-28
Retry if choco install winflexbison fails

ydah:
Sometimes `choco install winflexbison` fails, so how about retrying?
- https://github.com/ruby/lrama/actions/runs/7019310484/job/19096471447


yui-knk:
Thank you! I hope this will reduce errors on Windows CI.

PR #285, opened 2023-12-07
Support user define parameterizing rules

ydah:


parser.y
```diff
@@ -202,6 +203,73 @@ rule
...
+  rule_rhs_list: rule_rhs
+                {
+                  builder = val[0]
+                  result = [builder]
+                }
+          | rule_rhs_list "|" rule_rhs
+                {
+                  builder = val[2]
+                  result = val[0].append(builder)
+                }
```
yui-knk, line 224: [Note] Trailing `;` is covered by https://github.com/ruby/lrama/pull/285/files#diff-359cd0950b35838d2e90b9e88f5b91a34db81d3125073d530b3b4f4adbc3d0f2R72

lib/lrama/grammar/parameterizing_rule_resolver.rb
```diff
@@ -0,0 +1,31 @@
...
+      def add_parameterizing_rule_builder(builder)
+        @parameterizing_rule_builders << builder
+      end
+
+      def defined?(name)
+        @parameterizing_rule_builders.any? { |builder| builder.name == name }
+      end
+
+      def build_rules(token, rule_counter, lhs_tag, user_code, precedence_sym, line)
+        @parameterizing_rule_builders.each do |builder|
```
yui-knk, line 22: It may be needed to filter `@parameterizing_rule_builders` to only single builder whose name matches with `token.s_value` otherwise rules are generated for all `@parameterizing_rule_builders`.
Considering about the case below:

```
%rule defined_option(X): /* empty */
                       | X
                       ;

%rule defined_option2(X): /* empty */
                        | X
                        ;

%%

program         : defined_option(number) <i>
                ;
```

parser.y
```diff
@@ -202,6 +203,73 @@ rule
...
+
+  rule_rhs: /* empty */
```
yui-knk, line 226: [Note] Rule for nested parameterizing rules like `rhs IDENTIFIER parameterizing_suffix tag_opt` or `rhs IDENTIFIER "(" parameterizing_args ")" tag_opt` are not defined now. It will be supported by another PR.

lib/lrama/grammar/parameterizing_rule_builder.rb
```diff
@@ -0,0 +1,47 @@
...
+
+      def initialize(name, args, rhs)
+        @name = name
+        @args = args
+        @rhs = rhs
+        @term = nil
+        @required_args_count = args.count
+      end
+
+      def build_rules(token, build_token, rule_counter, lhs_tag, user_code, precedence_sym, line)
```
yui-knk, line 14: `user_code` is defined by `%rule` and can not be overridden by caller. Then I think `user_code` argument is not needed.

lib/lrama/grammar/parameterizing_rule_builder.rb
```diff
@@ -0,0 +1,47 @@
...
+        @rhs.each do |rhs|
+          @term = rhs_term(token, rhs)
+          rules << Rule.new(id: rule_counter.increment, _lhs: build_token, _rhs: [@term].compact, lhs_tag: lhs_tag, token_code: rhs.user_code, precedence_sym: rhs.precedence_sym, lineno: line)
+        end
+        rules
+      end
+
+      def build_token(token)
+        validate_argument_number!(token)
+        Lrama::Lexer::Token::Ident.new(s_value: "#{name}_#{token.args.first&.s_value}")
```
yui-knk, line 26: Generated rule name should include information about all arguments otherwise rule name conflict happens when
1. ParameterizingRule requires 2 or more arguments
2. Same ParameterizingRule is called with different arguments but their first argument is same.

For example,

```
%rule my_rule(X, Y) ...

%%

program: my_rule(number, number)
              | my_rule(number, string)
             ;
```

lib/lrama/grammar.rb
```diff
@@ -400,6 +408,10 @@ def normalize_rules
           add_nterm(id: rule._lhs)
         end
       end
+
+      if @parameterizing_resolver.term
+        add_term(id: @parameterizing_resolver.term)
```
yui-knk, line 413: I'm wondering this part, because
1. Only non terminals (nterm) are newly generated by applying parameterizing rules. In general terminals are defined by `%token` and non terminals are defined as LHS of rules. All of arguments are defined by `%token` as rules then LHS of newly generated rules are only new symbols we need to take care about here.
2. Non terminals we need to take care about is not always single. If 2 or more applications happen, 2 or more non terminals might be generated (exception is when all arguments of applications are same).

In short, this codes will be similar with https://github.com/ruby/lrama/pull/285/files#diff-bc2fdb878ce319ff339914b979ff298e61490736ab36ee1080bd8f0ff85b4a87R402-R405.

spec/fixtures/parameterizing_rules/user_defined.y
```diff
@@ -0,0 +1,42 @@
...
+%{
+// Prologue
+static int yylex(YYSTYPE *val, YYLTYPE *loc);
+static int yyerror(YYLTYPE *loc, const char *str);
+
+%}
+%union {
+    int i;
+}
+
```
yui-knk, line 14: If `number` is a token (terminal), it should be defined with `%token <i> number`, `<i>` is optional in this case.

lib/lrama/grammar.rb
yui-knk: I guess these codes are not needed. Nterm for new rules are covered by https://github.com/ruby/lrama/pull/285/files#diff-bc2fdb878ce319ff339914b979ff298e61490736ab36ee1080bd8f0ff85b4a87R402-R405.

lib/lrama/grammar/rule_builder.rb
```diff
@@ -110,11 +110,16 @@ def process_rhs
           when Lrama::Lexer::Token::Ident
             @replaced_rhs << token
           when Lrama::Lexer::Token::Parameterizing
-            parameterizing = ParameterizingRules::Builder.new(token, @rule_counter, @lhs_tag, user_code, precedence_sym, line)
-            parameterizing.build.each do |r|
-              @parameterizing_rules << r
+            if parameterizing_resolver.defined?(token.s_value)
+              parameterizing_resolver.build_rules(token, @rule_counter, @lhs_tag, line)
+              @parameterizing_rules = @parameterizing_rules + parameterizing_resolver.rules
```
yui-knk, line 115: `parameterizing_resolver.rules` stores all rules. Using `parameterizing_resolver.rules` in RuleBuilder causes multiple same rules.
`parameterizing_resolver.rules` is a cache for reusing same rules for same arguments, then `ParameterizingRuleResolver@rules` should be hidden from outside of class and `ParameterizingRuleResolver#build_rules` should return generated (or cached) rules corresponding to arguments.
For example

```
/*
 * This is comment for this file.
 */

%{
// Prologue
static int yylex(YYSTYPE *val, YYLTYPE *loc);
static int yyerror(YYLTYPE *loc, const char *str);

%}
%union {
    int i;
}

%token number

%rule defined_option(X): /* empty */
                       | X
                       ;

%rule defined_option2(X): /* empty */
                       | X
                       ;

%rule unused_define(X): /* empty */
                      | X
                      ;

%%

program         : defined_option(number) <i>
                | defined_option2(number) <i>
                ;

%%

static int yylex(YYSTYPE *yylval, YYLTYPE *loc) {
{
  return 0;
}

static int yyerror(YYLTYPE *loc, const char *str) {
{
  return 0;
}

int main(int argc, char *argv[])
{
}
```

```
$ bundle exec exe/lrama --trace=rules spec/fixtures/parameterizing_rules/user_defined.y

shift/reduce conflicts: 1 found
reduce/reduce conflicts: 1 found
Grammar rules:
$accept -> program, YYEOF
program -> defined_option_number
defined_option_number -> ε
defined_option_number -> number
program -> defined_option_number, defined_option2_number
defined_option_number -> ε
defined_option_number -> number
defined_option2_number -> ε
defined_option2_number -> number
```

lib/lrama/grammar/rule_builder.rb
```diff
@@ -110,11 +110,16 @@ def process_rhs
...
+              @replaced_rhs = @replaced_rhs + parameterizing_resolver.tokens
```
yui-knk, line 116: [nits] Is this same with `@replaced_rhs << parameterizing_resolver.tokens` ? If so I prefer `<<` so that it matches with other lines.

yui-knk:
Thanks for the PR. Direction is good.
I left comments, this functionality is big one, then we need to take care of a lot things. If you have any questions, ask me freely.
[Note] is just note for me, you don't need to care about in this PR.

lib/lrama/grammar/parameterizing_rule_resolver.rb
ydah: Thank you very much! The following commit fixes the problem:
https://github.com/ruby/lrama/pull/285/commits/b07ce82bef231fe02290ea480ee1371411c2c8d8

lib/lrama/grammar/parameterizing_rule_builder.rb
ydah: Indeed it is. Perhaps also `precedence_sym`. Removed in the following commit:
https://github.com/ruby/lrama/pull/285/commits/69b95e7cf6a75e4fbf80a73f3a3807cc841e605d

spec/fixtures/parameterizing_rules/user_defined.y
ydah: Oh... sorry I missed it 😣 Added in the following commit:
https://github.com/ruby/lrama/pull/285/commits/7ad441cf3b07deafd7fa316b336626d8f6d352e5

lib/lrama/grammar.rb
ydah: This is a needless process. It was caused by the same thing as https://github.com/ruby/lrama/pull/285#discussion_r1421240833. It is no longer needed and was removed in the following commit:
https://github.com/ruby/lrama/pull/285/commits/e8f022787f79e76fd8b2f0fdbbd10645fa6917a0

lib/lrama/grammar/rule_builder.rb
ydah: Indeed it is. Thanks for catching this! 
Fixed in the following commit:
- https://github.com/ruby/lrama/pull/285/commits/7add3f5780174f0881ab69bc83c26fcf1b7a6065

lib/lrama/grammar/rule_builder.rb
ydah: I believe the following modifications have been made together! Thanks🙏
https://github.com/ruby/lrama/pull/285#discussion_r1421707337

lib/lrama/grammar/parameterizing_rule_builder.rb
ydah: I had missed that. Thank you! I have fixed it in the following commit:
https://github.com/ruby/lrama/pull/285/commits/74f43ae67195b2befddd933ace329a1648499c51

ydah:
I updated this PR. Your detailed comments on the review allowed me to make corrections without question. Thank you!

yui-knk:
Thanks for the update, let's introduce user defined parameterizing rules into Lrama 🎉

PR #289, opened 2023-12-11
Integration test for user defined parameterizing rules

yui-knk:
There are two blockers to merge this test case.
1. `#yyr2` assumes rules are sorted by id. But it's not ensured when rules are generated by parameterizing rule. This was solved by https://github.com/ruby/lrama/pull/288.
2. `ParameterizingRuleRhsBuilder` only accept one symbol for rhs (`#symbol`). But it should accept arbitrary number of symbols. https://github.com/ruby/lrama/pull/291 will fix this issue.


yui-knk:
This was solved by https://github.com/ruby/lrama/pull/288 and https://github.com/ruby/lrama/pull/291.

PR #290, opened 2023-12-11
Use `parameterizing_args` instead of `rule_args`

ydah:


parser.y
```diff
@@ -203,15 +203,12 @@ rule
 
   token_declaration: id int_opt alias { result = val }
 
-  rule_declaration: "%rule" IDENTIFIER "(" rule_args ")" ":" rule_rhs_list
+  rule_declaration: "%rule" IDENTIFIER "(" parameterizing_args ")" ":" rule_rhs_list
```
yui-knk, line 206: `rule_args` is a list of `IDENTIFIER`. On the other hand `parameterizing_args` is a list of `symbol`. `symbol` means `IDENTIFIER`, `CHARACTER` or `STRING`.
In my understanding, using `symbol` for arguments make sense, but only `IDENTIFIER` makes sense for parameters.

ydah: Ah. Sorry. I was confused 🙇‍♂ You are right. Thank you.

yui-knk: No worries!

PR #291, opened 2023-12-11
Fix failed integration test for user defined parameterizing rules

ydah:
Follow up: https://github.com/ruby/lrama/pull/289

### DIFF
--- a/lib/lrama/grammar/parameterizing_rule_builder.rb
+++ b/lib/lrama/grammar/parameterizing_rule_builder.rb
@@ -15,7 +15,7 @@ module Lrama
         lhs = lhs_token(token)
         rules = []
         @rhs.each do |rhs|
-          rules << Rule.new(id: rule_counter.increment, _lhs: lhs, _rhs: [rhs_token(token, rhs)].compact, lhs_tag: lhs_tag, token_code: rhs.user_code, precedence_sym: rhs.precedence_sym, lineno: line)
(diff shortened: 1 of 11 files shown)


lib/lrama/grammar/parameterizing_rule_builder.rb
```diff
@@ -33,12 +33,10 @@ def lhs_token(token)
...
 
       def rhs_token(token, rhs)
-        return nil unless rhs.symbol
-        term = rhs.symbol
-        @args.each_with_index do |arg, index|
-          term = token.args[index] if arg.s_value == rhs.symbol.s_value
-        end
-        term
+        return [] if rhs.symbols.empty?
+        @args.each_with_index.map do |arg, i|
```
yui-knk, line 37: RHS of ParameterizingRule can contain other than parameters. For example `','` in below.

```
%rule pair(X, Y): X ',' Y { printf("(%d, %d)\n", $1, $2); }
                ;
```

Therefore this logic should be `map` over `rhs.symbols` with a block where symbol matches with an element of `@args` are replaced.
This also means line 36 can be removed.

yui-knk: In other words

```
%rule pair(X, Y): X ',' Y { printf("(%d, %d)\n", $1, $2); }
                ;

%%

stmt: pair(ODD, EVEN) { printf("pair odd even\n"); }
    ;
```

Should generate `pair_ODD_EVEN: ODD ',' EVEN`. But `@args.each_with_index.map` ignores `','` because `','` is stored in `rhs.symbols`.

yui-knk:
LGTM, thank you!

PR #327, opened 2023-12-28
Refactor for parameterizing rules

ydah:


lib/lrama/binding.rb
```diff
@@ -0,0 +1,23 @@
+module Lrama
+  class Binding
```
yui-knk, line 2: `Binding` is only used when building actual rules from grammar file, then I think it's better to be placed under `Grammar` namespace.

lib/lrama/grammar/rule_builder.rb
```diff
@@ -33,7 +34,7 @@ def add_rhs(rhs)
 
       def user_code=(user_code)
         if !@line
-          @line = user_code.line
+          @line = user_code&.line
```
yui-knk, line 37: [note] I will revisit what's the case where `user_code` doesn't have `line`.

yui-knk: Understand

```
lib/lrama/grammar/rule_builder.rb:123:41: [error] Cannot pass a value of type `(::Lrama::Lexer::Token::UserCode | nil)` as an argument of type `::Lrama::Lexer::Token::UserCode`
│   (::Lrama::Lexer::Token::UserCode | nil) <: ::Lrama::Lexer::Token::UserCode
│     nil <: ::Lrama::Lexer::Token::UserCode
│
│ Diagnostic ID: Ruby::ArgumentTypeMismatch
│
└                 rule_builder.user_code = r.user_code
                                           ~~~~~~~~~~~
```

lib/lrama/binding.rb
```diff
@@ -0,0 +1,23 @@
...
+      @actual_args = actual_args
+      @count = parameters.count
+      @parameter_to_arg = parameters.zip(actual_args).map do |param, arg|
+        [param.s_value, arg]
+      end.to_h
+    end
+
+    def resolve_symbol(symbol)
+      if symbol.is_a?(Lexer::Token::InstantiateRule)
+        symbol.args.map! { |arg| resolve_symbol(arg) }
```
yui-knk, line 16: Is it safe to destruct `InstantiateRule#args` when `InstantiateRule` appears in ParameterizingRule, so that the `InstantiateRule` might be referred more than once?

yui-knk: It might be better to build new `InstantiateRule` from `symbol` with replaced `args`.

yui-knk:
Direction is good, could you check comments without `[note]` ?

lib/lrama/binding.rb
ydah: Indeed it is. I changed the `InstantiateRule` to rebuild.

lib/lrama/binding.rb
ydah: Thank you. You are indeed correct, so I have updated this PR.

ydah:
Thank you for your review! I updated this PR.

yui-knk:
Go ahead!

PR #334, opened 2023-12-29
Resolve shadowing of local variables

ydah:


yui-knk:
Shadowing doesn't change the value of outer local variables, then it's not harmful. What is the motivation to avoid shadowing?

ydah:
The motivation was to make the variable readable by changing it to a variable with a different name than the outside local variable. As you say, I leave it to you to decide whether to merge or not, as there is no problem with the code as it stands now.

PR #336, opened 2023-12-31
Fix an infinite loop issue with recurrent calls to parameterizing rule

ydah:
This PR fixes the issue of an infinite loop when there is a recurrent parameterizing rule call.

For example, the following rule would have resulted in an infinite loop.

```ruby
%rule list(X): /* empty */
             | X
             | list(X)
             ;
```

### DIFF
--- a/lib/lrama/grammar/binding.rb
+++ b/lib/lrama/grammar/binding.rb
@@ -4,8 +4,6 @@ module Lrama
       attr_reader :actual_args, :count
 
       def initialize(parameterizing_rule, actual_args)
-        @rule_name = parameterizing_rule.name
-        @required_parameters_count = parameterizing_rule.required_parameters_count
         @parameters = parameterizing_rule.parameters
         @actual_args = actual_args
         @parameter_to_arg = @parameters.zip(actual_args).map do |param, arg|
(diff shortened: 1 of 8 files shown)


yui-knk:
I also want to resolve the issue that same rules are generated multiple times when a parameterizing rule is instantiated with same arguments.
For example file like below generates two `option_number`. This causes reduce/reduce conflicts because it is impossible to determine which `option_number` is used.

```
%union {
    int i;
}

%token number string

%rule option(Y) : /* empty */
                | Y
                ;

%%

program         : option(number)
                | string option(number)
                ;

%%
```

```shell
./exe/lrama --trace=rules tmp/sample.y
reduce/reduce conflicts: 3 found
Grammar rules:
$accept -> program, YYEOF
option_number -> ε
option_number -> number
program -> option_number
option_number -> ε
option_number -> number
program -> string, option_number
```

Both infinite loop and duplicated rules come from multiple instantiation of same parameterizing rule with same arguments. We can avoid it by recording the name of new rule's LHS and checking the name before building rules.
The records should be shared among all RuleBuilders, so that it might be an option to make `parameterizing_rule_resolver` to holds LHS names with hash. If you have better idea where it's recorded, please let me know.

ydah:
Thanks for the review. Changed to avoid instantiating the same parameterization rule multiple times with the same arguments.

```
❯ ./exe/lrama --trace=rules tmp/sample.y
Grammar rules:
$accept -> program, YYEOF
option_number -> ε
option_number -> number
program -> option_number
program -> string, option_number
```

yui-knk:
Thank you!

PR #344, opened 2024-01-11
Provide standard library as stdlib.y

ydah:
doc: https://gist.github.com/ydah/fda6f4abac4ce34547fe90143eba33f2


stdlib.y
```diff
@@ -0,0 +1,76 @@
...
+               ;
+
+/*
+ * program: list(number)
+ *
+ * =>
+ *
+ * program: list_number
+ * list_number: ε
+ * list_number: list_number number
```
yui-knk, line 31: `list_number -> number list_number`

stdlib.y
```diff
@@ -0,0 +1,76 @@
...
+             ;
+
+/*
+ * program: nonempty_list(number)
+ *
+ * =>
+ *
+ * program: nonempty_list_number
+ * nonempty_list_number: number
+ * nonempty_list_number: nonempty_list_number number
```
yui-knk, line 44: `nonempty_list_number -> number nonempty_list_number`

stdlib.y
```diff
@@ -0,0 +1,76 @@
...
+                      ;
+
+/*
+ * program: separated_nonempty_list(comma, number)
+ *
+ * =>
+ *
+ * program: separated_nonempty_list_comma_number
+ * separated_nonempty_list_comma_number: number
+ * separated_nonempty_list_comma_number: number, comma, separated_nonempty_list_comma_number
```
yui-knk, line 57: `,` is not needed on RHS.

stdlib.y
```diff
@@ -0,0 +1,76 @@
...
+ * program: separated_list(comma, number)
+ *
+ * =>
+ *
+ * program: separated_list_comma_number
+ * separated_list_comma_number: option_separated_nonempty_list_comma_number
+ * option_separated_nonempty_list_comma_number: ε
+ * option_separated_nonempty_list_comma_number: separated_nonempty_list_comma_number
+ * separated_nonempty_list_comma_number: number
+ * separated_nonempty_list_comma_number: number, comma, separated_nonempty_list_comma_number
```
yui-knk, line 73: Ditto. `,` is not needed.

yui-knk:
I agree to the direction to define standard library in "stdlib.y" file, it makes clear for us what's each standard rule means.
However I have concerns for current design.

## standard library by default or not

I think it's better to enable standard library by default. If some rules don't match with user's expectation, user can (or should be able to) overwrite the rule by themselves with writing same name parameterizing rule in their grammar file.

I aslo concern about the location where `%include-stdlib` can be written is very limited. As a user of parser generator, I don't want to take care of the location of each directive.

My suggestion is to make standard library by default.
In current design, grammar file parse and evaluation are separated. `Parser#do_parse` parsers grammar file and stores the structure into `Grammar` instance, `Grammar#prepare` is the method to instantiate rules. Both of them are called within `Parser#parse` but it's ok to move `@grammar.prepare` (and `@grammar.validate!`) outside `Parser#parse`.
We can merge the result of "stdlib.y" parse and grammar file parse before before instantiate rules.

## The location of "stdlib.y"

"stdlib.y" needs to be included into gem package and https://github.com/ruby/ruby/tree/master/tool/lrama. We can manage both of them by configuring "lrama.gemspec" and changing install script. But I feel it's good to put "stdlib.y" under "lib", e.g. "lib/lrama/grammar/".

stdlib.y
ydah: Oh sorry,  I remove `,`

stdlib.y
ydah: I remove `,` Thank you

ydah:
Thank you for the review. I agree that it should be enabled by default.
I have indeed made the modifications as I found that if I design it as you suggested, it will not be a problem.

The documentation has also been updated as per the current design.

lib/lrama/grammar/stdlib.y
```diff
@@ -0,0 +1,80 @@
...
+ * program: list(number)
+ *
+ * =>
+ *
+ * program: list_number
+ * list_number: ε
+ * list_number: number list_number
+ */
+%rule list(X): /* empty */
+             | X list(X)
```
yui-knk, line 34: The order of RHS tokens are changed, is this intentional?

lib/lrama/grammar/stdlib.y
```diff
@@ -0,0 +1,80 @@
...
+ * program: nonempty_list(number)
+ *
+ * =>
+ *
+ * program: nonempty_list_number
+ * nonempty_list_number: number
+ * nonempty_list_number: number nonempty_list_number
+ */
+%rule nonempty_list(X): X
+                      | X nonempty_list(X)
```
yui-knk, line 47: The order of RHS tokens are changed, is this intentional?

lib/lrama/grammar/stdlib.y
```diff
@@ -0,0 +1,80 @@
...
+ * program: separated_nonempty_list(comma, number)
+ *
+ * =>
+ *
+ * program: separated_nonempty_list_comma_number
+ * separated_nonempty_list_comma_number: number
+ * separated_nonempty_list_comma_number: number comma separated_nonempty_list_comma_number
+ */
+%rule separated_nonempty_list(separator, X): X
+                                           | X separator separated_nonempty_list(separator, X)
```
yui-knk, line 60: The order of RHS tokens are changed, is this intentional?

lib/lrama/option_parser.rb
```diff
@@ -59,6 +59,7 @@ def parse_by_option_parser(argv)
         o.on('-S', '--skeleton=FILE', 'specify the skeleton to use') {|v| @options.skeleton = v }
         o.on('-t', 'reserved, do nothing') { }
         o.on('--debug', 'display debugging outputs of internal parser') {|v| @options.debug = true }
+        o.on('--no-stdlib', 'do not include standard library') {|v| @options.no_stdlib = true }
```
yui-knk, line 62: I prefer grammar file declaration like `%no-stdlib` to command line options because grammar file declaration is more declarative than command line option. It's difficult to track what options are passed from parse.y file.
What do you think about it?

yui-knk:
Left some comments, others seems good to me!

lib/lrama/grammar/stdlib.y
ydah: I changed it to be the same as the order of Menhir, but would the original order be better?
https://gitlab.inria.fr/fpottier/menhir/-/blob/20230608/src/standard.mly?ref_type=tags#L209

lib/lrama/grammar/stdlib.y
ydah: Ditto. https://gitlab.inria.fr/fpottier/menhir/-/blob/20230608/src/standard.mly?ref_type=tags#L219

lib/lrama/grammar/stdlib.y
ydah: Ditto. https://gitlab.inria.fr/fpottier/menhir/-/blob/20230608/src/standard.mly?ref_type=tags#L239

lib/lrama/option_parser.rb
ydah: It would certainly be better that way. I agree with your opinion. I will change it.

ydah:
Thank you for the review! I have updated this PR.

lib/lrama/grammar/stdlib.y
yui-knk: It's interesting. I guess Menhir's rule comes from the fact that new item is appended on the head of list in OCaml, see `(::)` in https://v2.ocaml.org/api/List.html.
On the other hand, new item is appended on the last of the list in Ruby parse.y, for example https://github.com/ruby/ruby/blob/7cf74a2ff28b1b4c26e367d0d67521f7e1fed239/parse.y#L4132.
Therefore I think the original order is better.

ydah: I see. Thank you for explaining! I changed to use the original order.

ydah:
Thank you for the review! I have updated this PR.

yui-knk:
Excellent job 💯

PR #349, opened 2024-01-17
`RuleBuilder#numberize_references` should check LHS and RHS for duplication

yui-knk:
Before this commit, LHS is not included into check list of `candidates`. Therefore `$args` in the code below is valid reference and converted to the reference to LHS (`args`).
This commit change it to check both of LHS and RHS when resolve the named reference, so that the code is reported as duplicated reference.

```
args : ...
     | args ',' arg_splat
         {
             $$ = rest_arg_append(p, $args, $arg_splat, &@$);
         }
     ;
```

### DIFF
--- a/lib/lrama/grammar/rule_builder.rb
+++ b/lib/lrama/grammar/rule_builder.rb
@@ -173,11 +173,12 @@ module Lrama
 
           token.references.each do |ref|
             ref_name = ref.name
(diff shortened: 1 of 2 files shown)


yui-knk:
https://github.com/ruby/ruby/pull/9634 was merged then I merge this PR.

PR #353, opened 2024-01-23
Bump min ruby version from 2.5 to 2.7 in GHA

ydah:


yui-knk:
Thanks for the PR. However at the moment I want to keep Ruby 2.5 as a minimal BASERUBY version.
Because Ruby 3.3 released by 2023-12-25 is the first Ruby version which uses Lrama. Then I'm still wondering what is the best practice for maintaining Lrama for ruby stable branches. I plan to use Lrama v0_5 for Ruby 3.3 however if v0_6 (master branch) works for both Ruby 3.3 and 3.4, we have an option to use Lrama v0_6 in Ruby 3.3 branch.
Therefore I want to keep minimal BASERUBY at the moment.

ydah:
Thank you. I agree with your opinion. If needed, let's open a new pull request again. For now, I will go ahead and close this pull request.

PR #364, opened 2024-01-28
Refactoring to extract Lrama::Grammar::Symbols

ydah:


yui-knk:
Thanks for the PR. I think separating Symbol generation code into `Resolver` makes `Grammar` clean. However class hierarchy of `Symbols::Base`, `Symbols::Nterms` and `Symbols::Terms` is difficult. And accessing `terms` and `nterms` with `@states.terms.symbols` or `@states.nterms.symbols` is not intuitive because in this context `symbol` is a set of `terms` and `nterms`.

Can I ask you to
1. Move all Symbol generation code, including `#fill_symbol_number` and so on, into `Resolver`
2. Remove `Symbols::Base`, `Symbols::Nterms` and `Symbols::Terms`
3. Keep `Grammar#terms` and `Grammar#tnerms` both of them will be delegated to `symbols_resolver`

ydah:
Fix to remove Symbols:: Base, Symbols:: Nterms, and Symbols:: Terms to delegate processing to resolvers. What do you think? Please point out if I missed something.

lib/lrama/grammar.rb
```diff
@@ -273,7 +195,7 @@ def compute_nullable
         end
 
         rule_count_2  = @rules.count {|e| e.nullable.nil? }
-        nterm_count_2 = nterms.count {|e| e.nullable.nil? }
+        nterm_count_2 = nterms.select {|e| e.nullable.nil? }.count
```
yui-knk, line 198: I overlooked this change in previous review. Could you keep the original codes? This line and previous line execute same things for other sets. So that it's good to be same pattern.

yui-knk:
Left one comment, other seems good.

lib/lrama/grammar.rb
ydah: Thank you catching this. I reverted this code.

yui-knk:
Thanks!

PR #366, opened 2024-01-31
chore: trivial refactoring

ydah:


yui-knk:
A lot of changes looks good. Two points I want to keep are:
1. `<<-` style for "lib/lrama/output.rb" https://github.com/ruby/lrama/pull/366/commits/786330f8978e413e489c5b51b03019b558a2839f because they are embedded C codes then it's easy for me to understand what actually embedded by using `<<-`
2. Previous codes of https://github.com/ruby/lrama/pull/366/commits/bc89c4e8babdffbf578dce5d3f4291fdf4a4e5c6 because currently I feel it's simple to modify or print something in if block when it's one line, this is just my preference.

ydah:
Sorry for the delay, I changed it back to <<- for lib/lrama/output.rb. Also, I removed the bc89c4e change.

yui-knk:
Thank you!

PR #369, opened 2024-02-09
Introduce Inlining

ydah:
This PR implements the [Inlining](https://gallium.inria.fr/~fpottier/menhir/manual.html#sec37) feature in Menhir in Lrama.


parser.y
```diff
@@ -30,6 +30,7 @@ rule
 
   bison_declaration: grammar_declaration
                    | rule_declaration
+                   | inline_declarations
```
yui-knk, line 33: I guess `%inline` is an optional attribute for `%rule` then its production rules are declared as a part of `rule_declaration` like:

```
rule_declaration: "%rule" inline_opt IDENTIFIER "(" rule_args ")" ":" rule_rhs_list`

inline_opt: # empty
          | "%inline"           
```

ydah: Indeed it is. I changed `%inline` to an optional attribute of `%rule` https://github.com/ruby/lrama/pull/369/commits/0cf305a2d46ec98a804943ee0d259e23762d7258

parser.y
```diff
@@ -237,6 +238,12 @@ rule
                         @grammar.add_parameterizing_rule(rule)
                       }
 
+  inline_declaration: "%rule" "%inline" id_colon ":" rule_rhs_list
```
yui-knk, line 241: Just for clarification, this grammar rule doesn't allow parameter for inline rule. For example, "%rule %inline op(a): ..." is not allowed.
Is this intentional?
If it will be supported in following PR, it's ok for me.

ydah: Yes. I plan to implement it in a follow-up PR. Sorry for the confusion :bow:

lib/lrama/grammar/parameterizing_rule/resolver.rb
```diff
@@ -13,8 +13,14 @@ def add_parameterizing_rule(rule)
           @rules << rule
         end
 
-        def find(token)
-          select_rules(token).last
+        def find_rule(token)
+          if token.is_a?(Lexer::Token::InstantiateRule)
```
yui-knk, line 17: Is there any reason for this check?
As far as I read the code, caller of this method (https://github.com/ruby/lrama/blob/b08572c2cf4ce0672ffc99d305583362f05f6154/lib/lrama/grammar/rule_builder.rb#L120) ensures `token` is a `InstantiateRule`. If there is not reason to return `nil` for other than `InstantiateRule`, it's better to raise an error whose message includes `BUG`.

lib/lrama/grammar/parameterizing_rule/resolver.rb
```diff
@@ -25,7 +31,7 @@ def created_lhs(lhs_s_value)
 
         def select_rules(token)
           rules = select_rules_by_name(token.rule_name)
-          rules = rules.select { |rule| rule.required_parameters_count == token.args_count }
+          rules = rules.select { |rule| rule.required_parameters_count == token.args_count && !rule.is_inline }
```
yui-knk, line 34: [nits] It might be better to extract `rules.select {|rule| !rule.is_inline}` into another method.

lib/lrama/grammar/rule_builder.rb
```diff
@@ -52,12 +57,16 @@ def complete_input
 
       def setup_rules(parameterizing_rule_resolver)
         preprocess_references unless @skip_preprocess_references
-        process_rhs(parameterizing_rule_resolver)
+        if rhs.any? { |token| parameterizing_rule_resolver.find_inline(token) }
```
yui-knk, line 60: [note] In the future, we need to change this condition otherwise lrama can not handle production rules whose RHS include both parameterizing rules and inline rules.

parser.y
yui-knk: Understand, no worries.

yui-knk:
@ydah Reviewed. The direction is ok for me.

* https://github.com/ruby/lrama/pull/369#discussion_r1545960124 could you check this?
* [nits] Could you check it?
* [note] is just note for me. No update is needed right now

yui-knk:
This PR is the first step to support inline rule in Lrama. Then only support primitive functionality, for example nested inline rules and combination of inline rules and non inline rules are not supported now.

lib/lrama/grammar/parameterizing_rule/resolver.rb
ydah: Indeed it is! Thank you. I updated this PR: https://github.com/ruby/lrama/pull/369/commits/bba8995d6222091cda48b69663e4a6defac6e87b

lib/lrama/grammar/parameterizing_rule/resolver.rb
ydah: That's good. I updated this PR: https://github.com/ruby/lrama/pull/369/commits/6928bc856a5358195533ed13bb021f6e6f980ef5

ydah:
@yui-knk Thank you for your review! I updated this PR.

lib/lrama/grammar/parameterizing_rule/resolver.rb
```diff
@@ -13,17 +13,22 @@ def add_parameterizing_rule(rule)
...
+
         def created_lhs(lhs_s_value)
           @created_lhs_list.reverse.find { |created_lhs| created_lhs.s_value == lhs_s_value }
         end
 
         private
 
         def select_rules(token)
+          rules = select_not_inline_rules
           rules = select_rules_by_name(token.rule_name)
```
yui-knk, line 32: It seems the result o line 31 is not used by line 32 and overwrote by line 32. Is something wrong?

ydah: 😨 My apologies... I updated it.

yui-knk:
Thank you!

PR #381, opened 2024-02-27
Fix an issue in a user-defined parameterizing rule where the caller cannot access `$$` in the rule

ydah:
For example, if `printf("pair odd even:% d\n," $1); `was unable to access the pair side of` $$`with` $1`, fix this:
```
%rule pair(X, Y): X Y
  {
    $$ = $1 + $2;
    printf("(%d, %d)\n", $1, $2);
  }
  ;

stmt: pair(ODD, EVEN) <num> { printf("pair odd even: %d\n", $1); }
```


yui-knk:
Thank you!

PR #391, opened 2024-03-24
Add support for Parameterizing rules with zero parameters

ydah:
In some cases, Inline does not specify parameters. Inline is implemented as an option to Parameterizing rules. 
Therefore, we added support for Parameterizing rules with zero parameters.

### DIFF
--- a/parser.y
+++ b/parser.y
@@ -231,7 +231,12 @@ rule
 
   token_declaration: id int_opt alias { result = val }
 
-  rule_declaration: "%rule" IDENTIFIER "(" rule_args ")" ":" rule_rhs_list
+  rule_declaration: "%rule" id_colon ":" rule_rhs_list
+                      {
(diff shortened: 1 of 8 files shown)


parser.y
```diff
@@ -231,7 +231,12 @@ rule
 
   token_declaration: id int_opt alias { result = val }
 
-  rule_declaration: "%rule" IDENTIFIER "(" rule_args ")" ":" rule_rhs_list
+  rule_declaration: "%rule" id_colon ":" rule_rhs_list
```
yui-knk, line 234: I think `%rule` **without** `%inline` is needless because it is same with 
a normal production rule.
I agree to introducing zero parameters for `%inline` but it might be better to support the feature in https://github.com/ruby/lrama/pull/369.

ydah: Indeed it is.  So I close this PR. Thank you for your review!

PR #394, opened 2024-04-09
Add support for adjusting Index to Inline

ydah:


yui-knk:
[note]
Some methods on `RuleBuilder` require `parameterizing_rule_resolver` for its argumen, e.g. `#setup_rules`, `#has_inline_rules?` and `#resolve_inline_rules`. It might be better to pass `parameterizing_rule_resolver` for  `RuleBuilder#initialize` by adding helper function to `Grammar` which calls `Grammar::RuleBuilder.new` with `@parameterizing_rule_resolver`
It's not directly related to this PR, so I don't require to change it in this PR.

lib/lrama/grammar/rule_builder.rb
```diff
@@ -204,6 +207,9 @@ def replace_inline_user_code(inline_rhs, index)
         return user_code if user_code.nil?
 
         code = user_code.s_value.gsub(/\$#{index + 1}/, inline_rhs.user_code.s_value)
+        user_code.references[(index+1)..-1]&.each do |ref|
```
yui-knk, line 210: I'm wondering several points for this method.

## 1. Order of `user_code.references`

The order of `user_code.references` is the order of reference appearance. For example if the code is `$3 $2 $1 $$`, the order of `user_code.references` is also `[$3, $2, $1, $$]`. This means `user_code.references[(index+1)..-1]` might not what we want to do. I think it will be like:

```ruby
user_code.references.each do |ref|
  next if ref.index.nil? || ref.index <= index # nil is a case for `$$`
  ...
end
```

## 2. Type of reference

`user_code.references` includes both `$n` and `@n`, we might need to take care of `@n` in the future

## 3. How to overwrite references

Right now this code overwrite code string, however another approach is to dup the code and overwrite reference index like below.
Manipulating references is helpful if we handle named references.

```ruby
        result = user_code.dup
        result.references[(index+1)..-1]&.each do |ref|
          ref.index += inline_rhs.symbols.count-1
        end
        result
```

yui-knk:
@ydah The direction seems good. Could you check the logic of reference overwrite, https://github.com/ruby/lrama/pull/394#discussion_r1563749823 ?

lib/lrama/grammar/rule_builder.rb
ydah: Thank you for your review!
I fixed 1 and 2 and added a commit.

However, for 3, I changed the way to write references overwrites, but I tried rewriting it like this:. However, when I wrote a test to check UserCode references in parser.spec, the index didn't seem to change.

```ruby
def replace_inline_user_code(inline_rhs, index)
  return user_code if inline_rhs.user_code.nil?
  return user_code if user_code.nil?

  code = user_code.s_value.gsub(/\$#{index + 1}/, inline_rhs.user_code.s_value)
  replaced_code = Lrama::Lexer::Token::UserCode.new(s_value: code, location: user_code.location)

  replaced_code.references.each do |ref|
    next if ref.index.nil? || ref.index <= index # nil is a case for `$$`

    ref.index += inline_rhs.symbols.count
  end
  replaced_code
end
```

ydah: If I'm missing something, please point it out 🙏

ydah:
@yui-knk Thank you for your review! I updated this PR.

ydah:
@yui-knk I rebased and fix conflict.

yui-knk:
LGTM!

PR #395, opened 2024-04-10
Refactor State#xxx_transitions

junk0612:
I was reading the State class and saw a place where it looked like it could be refactored.


yui-knk (approved):

yui-knk:
LGTM, thank you!

PR #396, opened 2024-04-17
Introduce the `--trace=actions` option

ydah:
The concept of merging actions was born in Inline. I think it would be useful to have an option to print the resolved action. However, we don't know what it is when I just display the action, so I've decided to print the rule and its corresponding action like this:. What do you think?

```
❯ exe/lrama --trace=actions sample/calc.y
Grammar rules with actions:
$accept -> list, YYEOF {}
list -> ε {}
list -> list, LF {}
list -> list, expr, LF { printf("=> %d\n", $2); }
expr -> NUM {}
expr -> expr, '+', expr { $$ = $1  +  $3; }
expr -> expr, '-', expr { $$ = $1  -  $3; }
expr -> expr, '*', expr { $$ = $1  *  $3; }
expr -> expr, '/', expr { $$ = $1  /  $3; }
expr -> '(', expr, ')' { $$ = $2; }
```

### DIFF
--- a/lib/lrama/command.rb
+++ b/lib/lrama/command.rb
@@ -47,6 +47,11 @@ module Lrama
         puts grammar.rules
       end
 
+      if options.trace_opts && options.trace_opts[:actions]
+        puts "Grammar rules with actions:"
+        grammar.rules.each { |rule| puts rule.with_actions }
+      end
+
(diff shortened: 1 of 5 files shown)


yui-knk:
LGTM, thank you!

PR #398, opened 2024-04-23
Support IELR(1) Parser Generation

junk0612:
Support the generation of the IELR(1) parser described in this paper.
https://www.sciencedirect.com/science/article/pii/S0167642309001191

This PR will be ready for merging when the following is done.

- [x] Add tests
    - Currently, I have only created a grammar file and verified its operation in my local environment.
- [x] Refactor
    - The data structure is awful and needs to be improved because the priority was to achieve the expected results.

### DIFF
--- a/lib/lrama/state.rb
+++ b/lib/lrama/state.rb
@@ -10,7 +10,7 @@ module Lrama
   class State
     attr_reader :id, :accessing_symbol, :kernels, :conflicts, :resolved_conflicts,
                 :default_reduction_rule, :closure, :items
-    attr_accessor :shifts, :reduces
+    attr_accessor :shifts, :reduces, :ielr_isocores, :lalr_isocore
 
     def initialize(id, accessing_symbol, kernels)
       @id = id
@@ -23,6 +23,9 @@ module Lrama
       @conflicts = []
       @resolved_conflicts = []
       @default_reduction_rule = nil
+      @predecessors = []
+      @lalr_isocore = self
+      @ielr_isocores = [self]
     end
 
     def closure=(closure)
@@ -84,6 +87,18 @@ module Lrama
       @transitions ||= shifts.map {|shift| [shift, @items_to_state[shift.next_items]] }
     end
 
+    def update_transition(shift, next_state)
+      set_items_to_state(shift.next_items, next_state)
+      next_state.append_predecessor(self)
+      clear_transitions_cache
+    end
+
+    def clear_transitions_cache
+      @nterm_transitions = nil
+      @term_transitions = nil
+      @transitions = nil
+    end
+
     def selected_term_transitions
       term_transitions.reject do |shift, next_state|
         shift.not_selected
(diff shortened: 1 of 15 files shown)


jasone:
A couple months ago I poked around the internet looking for independent IELR(1) implementations, and besides `lrama` found only an inactive [rust-based attempt](https://gitlab.com/arnaudgolfouse/ielr). Recently I completed my own implementation and wrote a [technical report](https://branchtaken.com/reports/ielr1/ielr1.html) with the hope that it might help others. IELR(1) turns out to be a real bear to figure out, even if it takes only a moderate amount of code. I hope you find the time and inspiration to complete this effort. Good luck!

junk0612:
On https://github.com/ruby/lrama/pull/398/commits/7da96768fb91d6e017102a0b9a06528d7cf43f39 , Lrama can parse CRuby's parse.y at https://github.com/ruby/ruby/commit/028958f9ad83adbd1f7cb8e79167c9fb5e694535 .

I want to add some files of results of parsing.
`ruby path/to/id2token path/to/parse.y | bundle exec lrama -v - lalr`
[lalr.output](https://github.com/user-attachments/files/17743319/lalr.output.txt)

`ruby path/to/id2token path/to/parse.y | bundle exec lrama -D lr.type=ielr -v - ielr`
[ielr.output](https://github.com/user-attachments/files/17743545/ielr.output.txt)

lib/lrama/states.rb
```diff
@@ -92,6 +92,20 @@ def compute
       report_duration(:compute_default_reduction) { compute_default_reduction }
     end
 
+    def compute_ielr
```
yui-knk, line 95: [note]: Ref "3.5.4. Algorithm" of the paper.

lib/lrama/states.rb
```diff
@@ -524,5 +541,52 @@ def compute_default_reduction
         end.first
       end
     end
+
+    def split_states
+      transition_queue = []
```
yui-knk, line 546: nits: It seems this variable is not used.

lib/lrama/states.rb
```diff
@@ -524,5 +541,52 @@ def compute_default_reduction
         end.first
       end
     end
+
+    def split_states
```
yui-knk, line 545: [note]: "Definition 3.45 (split_states)"

lib/lrama/state.rb
```diff
@@ -142,5 +160,274 @@ def rr_conflicts
         conflict.type == :reduce_reduce
       end
     end
+
+    def propagate_lookaheads(next_state)
```
yui-knk, line 164: [note]: Definition 3.40 (propagate_lookaheads)

lib/lrama/state.rb
```diff
@@ -142,5 +160,274 @@ def rr_conflicts
...
+      next_state.kernels.map {|item|
+        lookahead_sets =
+          if item.position == 1
```
yui-knk, line 167: [note]: Corresponding to case 2 of Definition 3.40 because `position` is 0-origin in this code base.

yui-knk, line 167: [nits]: I personally like to make the condition order to be aligned with the original paper for readability of the algorithm. For example 

```ruby
if item.position > 1
  ...
else
  ...
end
```

lib/lrama/state.rb
```diff
@@ -142,5 +160,274 @@ def rr_conflicts
...
+            goto_follow_set(item.lhs)
+          else
+            kernel = kernels.find {|k| k.predecessor_item_of?(item) }
```
yui-knk, line 170: [note]: `C[k]` is the kernel (1) whose LHS and RHS are same with the item's one and (2) whose position is `item.position - 1`.

lib/lrama/state.rb
```diff
@@ -142,5 +160,274 @@ def rr_conflicts
...
+            item_lookahead_set[kernel]
+          end
+
+        [item, lookahead_sets & next_state.lookahead_set_filters[item]]
```
yui-knk, line 174: [note]: This is same with the definition case 1 because case 1 requires tokens which are in both lookahead_set_filters of `s'` kernel and item_lookahead_sets of `s` kernel.

lib/lrama/states.rb
```diff
@@ -92,6 +92,20 @@ def compute
       report_duration(:compute_default_reduction) { compute_default_reduction }
     end
 
+    def compute_ielr
+      report_duration(:split_states) { split_states }
+      report_duration(:compute_direct_read_sets) { compute_direct_read_sets }
```
yui-knk, line 97: [note]: Compute LA set again, once states are splitted. See Phase 4 and Phase 5 on "3.1. Overview".

lib/lrama/state.rb
```diff
@@ -142,5 +160,274 @@ def rr_conflicts
...
+      }.to_h
+    end
+
+    def lookaheads_recomputed
+      !@item_lookahead_set.nil?
+    end
+
+    def compatible_lookahead?(filtered_lookahead)
```
yui-knk, line 182: [note]: See "Definition 3.43 (is_compatible)".

yui-knk: [nits]: If so, I prefer to use `is_compatible?` as a method name for clarity.

lib/lrama/state.rb
```diff
@@ -23,6 +23,12 @@ def initialize(id, accessing_symbol, kernels)
       @conflicts = []
       @resolved_conflicts = []
       @default_reduction_rule = nil
+      @predecessors = []
+      @lalr_isocore = self
```
yui-knk, line 27: [note]: 

* `@lalr_isocore` is an implementation of `lalr1_isocores` (Definition 3.36)
* `@ielr_isocores` is an implementation of `isocore_nexts` (Definition 3.37)
* Their initialization logic is written in Definition 3.45 (split_states).

lib/lrama/state.rb
```diff
@@ -142,5 +160,274 @@ def rr_conflicts
...
+    def compatible_lookahead?(filtered_lookahead)
+      !lookaheads_recomputed ||
+        @lalr_isocore.annotation_list.all? {|token, actions|
+          a = dominant_contribution(token, actions, item_lookahead_set)
+          b = dominant_contribution(token, actions, filtered_lookahead)
+          a.nil? || b.nil? || a == b
+        }
+    end
+
+    def lookahead_set_filters
```
yui-knk, line 191: [note]: Definition 3.38 (lookahead_set_filters)

lib/lrama/state.rb
```diff
@@ -142,5 +160,274 @@ def rr_conflicts
...
+    end
+
+    def annotate_manifestation
+      inadequacy_list.transform_values {|actions|
+        actions.map {|action|
+          if action.is_a?(Shift)
+            [action, nil]
+          elsif action.is_a?(Reduce)
+            if action.rule.empty_rule?
+              [action, lhs_contributions(action.rule.lhs, inadequacy_list.key(actions))]
```
yui-knk, line 266: [nits]: If need to get the key, I like to use `inadequacy_list.map` or `h = {}; inadequacy_list.each ...` so that the code accesses its key clearly.

lib/lrama/state.rb
```diff
@@ -142,5 +160,274 @@ def rr_conflicts
...
+              [action, contributions]
+            end
+          end
+        }.to_h
+      }
+    end
+
+    def annotate_predecessor(predecessor)
+      annotation_list.transform_values {|actions|
+        token = annotation_list.key(actions)
```
yui-knk, line 278: [nits]: Same with `#annotate_manifestation`, it's more clear to use `inadequacy_list.map` or `h = {}; inadequacy_list.each ...` to access its key.

lib/lrama/state.rb
```diff
@@ -142,5 +160,274 @@ def rr_conflicts
...
+      a.reject {|action|
+        if action.is_a?(State::Shift)
+          action.not_selected
+        elsif action.is_a?(State::Reduce)
+          action.not_selected_symbols.include?(token)
+        end
+      }
+    end
+
+    def inadequacy_list
```
yui-knk, line 217: [note]: Definition 3.27 (inadequacy_lists). Difference from the original paper is that tuple `(state, token, actions)` is used in the original paper but tuple `(token, actions)` is used in this implementation.

lib/lrama/state.rb
```diff
@@ -142,5 +160,274 @@ def rr_conflicts
...
+            end
+          }
+        }
+    end
+
+    def annotate_manifestation
+      inadequacy_list.transform_values {|actions|
+        actions.map {|action|
+          if action.is_a?(Shift)
+            [action, nil]
```
yui-knk, line 263: [note]: `nil` means "undefined" in the original paper.

lib/lrama/state.rb
```diff
@@ -142,5 +160,274 @@ def rr_conflicts
...
+          elsif action.is_a?(Reduce)
+            if action.rule.empty_rule?
```
yui-knk, line 265: [nits]: I prefer the order of sub condition follows the same order of the original paper.

lib/lrama/state.rb
```diff
@@ -142,5 +160,274 @@ def rr_conflicts
...
+              [action, lhs_contributions(action.rule.lhs, inadequacy_list.key(actions))]
+            else
+              contributions = kernels.map {|kernel| [kernel, kernel.rule == action.rule && kernel.end_of_rule?] }.to_h
```
yui-knk, line 268: [note]: "C[j] = (l -> r, |r| + 1)" means `end_of_rule`.

lib/lrama/state.rb
```diff
@@ -142,5 +160,274 @@ def rr_conflicts
...
+                  k.position == 1 && predecessor.lhs_contributions(k.lhs, token)[pred_k]
+                )
+              }]
+            }.to_h
+          end
+        }
+      }
+    end
+
+    def lhs_contributions(sym, token)
```
yui-knk, line 300: [note]: `sym` is the symbol (non-terminal) of the rule's LHS.

lib/lrama/state.rb
```diff
@@ -142,5 +160,274 @@ def rr_conflicts
...
+      shift, next_state = nterm_transitions.find {|sh, _| sh.next_sym == sym }
```
yui-knk, line 301: [nits]: Nonterminal transition is called goto. Shift is used for terminal transition.

spec/fixtures/integration/ielr.y
```diff
@@ -0,0 +1,62 @@
...
+%union {
+    int val;
+}
+
+%token a
+%token b
+%token c
+%define lr.type ielr
+
+%%
```
yui-knk, line 19: This grammar example comes from Fig 5 of the original paper. Could clarify it as comments?

yui-knk:
@junk0612 I created 2 test cases as reproduction of https://github.com/ruby/lrama/pull/398#issuecomment-2475464104
https://github.com/yui-knk/lrama/commit/ca288c87b79592826bc0aa7861aff963b002deb5

lib/lrama/states.rb
```diff
@@ -524,5 +541,52 @@ def compute_default_reduction
         end.first
       end
     end
+
+    def split_states
+      transition_queue = []
+      @states.each do |state|
```
yui-knk, line 547: I'm wondering it needs to limit the iteration only for the LALR states like `@states[0...@states.size].each ...`, in other words do we need to iterate states which are created by `#compute_state`?

yui-knk: It might be my misunderstandings. We may need to iterate all states which includes newly created one.

yui-knk (approved):

PR #399, opened 2024-04-26
Fix an error for using parameterizing rules with own stack in semantic action

ydah:
This PR fixes error.

```
tool/lrama/lib/lrama/grammar/code/rule_action.rb:58:in 'Integer#+': nil can't be coerced into Integer (TypeError)
```

### DIFF
--- a/spec/fixtures/integration/user_defined_parameterizing_rules.y
+++ b/spec/fixtures/integration/user_defined_parameterizing_rules.y
@@ -25,6 +25,7 @@ static int yyerror(YYLTYPE *loc, const char *str);
                     {
                         $$ = $1 + $2;
                         printf("(%d, %d)\n", $1, $2);
+                        printf("(%d, %d)\n", $:1, $:2);
                     }
                 ;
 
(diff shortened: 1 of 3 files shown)


spec/fixtures/integration/user_defined_parameterizing_rules.y
```diff
@@ -25,6 +25,7 @@ static int yyerror(YYLTYPE *loc, const char *str);
                     {
                         $$ = $1 + $2;
                         printf("(%d, %d)\n", $1, $2);
```
yui-knk, line 27: [note] `$n` could be used with `skip_preprocess_references: true` because `Reference#index` is set by `#scan_reference` then no need to fill index by `#numberize_references` (`#preprocess_references`) https://github.com/ruby/lrama/blob/3b1ade0a6584db75d5800cdcd67afe7e950a4b37/lib/lrama/lexer/token/user_code.rb#L43

yui-knk:
LGTM! Thank you!

PR #407, opened 2024-05-01
Add support for Named References for actions of rhs in Parameterizing rules

ydah:


lib/lrama/grammar/parameterizing_rule/rhs.rb
```diff
@@ -9,6 +9,19 @@ def initialize
...
         end
+
+        def resolve_user_code(bindings)
+          return unless user_code
+
+          code = user_code.s_value
+          symbols.each do |sym|
+            resolved_sym = bindings.resolve_symbol(sym)
+            if resolved_sym != sym
+              code = code.gsub(/\$#{sym.s_value}/, "$#{resolved_sym.s_value}")
```
yui-knk, line 20: Can we reuse `numberize_references`?

ydah: I think it is difficult to reuse.
The `reference` is creating in the following:
https://github.com/ruby/lrama/blob/f55f6a5e1521f837d4529cef4026b489e90669bb/lib/lrama/grammar/rule_builder.rb#L45

At this point, the s_value of Lexer:: Token:: UserCode must be the expanded value of parameterizing rules, as follows:
In other words, for a definition like this:

```
%rule foo(X): X { $$ = $X }

%%

bar: foo(baz)
       ;
```

RHS's X be baz and `$X` must also be` $baz`.
Am I missing something? In that case, please point it out.

spec/fixtures/parameterizing_rules/user_defined/with_action_and_named_references.y
```diff
@@ -0,0 +1,41 @@
...
+
+%union {
+    int i;
+    char *s;
+}
+
+%token <i> number
+%token <s> string
+
+%rule pair(X, Y): X ',' Y { printf("(%d, %d)\n", $X, $2); }
```
yui-knk, line 19: By the way, `$2` might be `$3`?

lib/lrama/grammar/parameterizing_rule/rhs.rb
yui-knk: I think two options.

## 1. Apply `#numberize_references` like method in `ParameterizingRule::Rule`

https://github.com/yui-knk/lrama/tree/parameterizing-meets-named-refs_option_1

This is a kind of porting of `#numberize_references`

## 2. Change `Reference#name` in `#resolve_user_code`

https://github.com/yui-knk/lrama/tree/parameterizing-meets-named-refs_option_2

This approach replace variable name with actual name.
For example:

```
%rule foo(X): X { $$ = $X }

%%

bar: foo(baz)
       ;
```

is replaced with `foo_bar: bar { $$ = $bar }`

## Which is better?

I prefer option 2 because

1. Option 2 has no duplicated logic. The structure of `RuleBuilder` and `ParameterizingRule::Rhs` is different. For example, `ParameterizingRule::Rhs` doesn't have LHS but `RuleBuilder` has.
2. Current implementation delegate `#preprocess_references` to `RuleBuilder` ([here](https://github.com/ruby/lrama/blob/236c0f86fc20f3f48f6cb32071382b712ed7cd0b/lib/lrama/grammar/rule_builder.rb#L140)). Option 2 respects current implementation.

In any case, I want to manipulate `Reference` rather than `UserCode#s_value` because `Reference` is more structured than `UserCode#s_value`.

ydah: Thank you for showing that. I also agree that Option2 is better. I cherry-picked your commit and deleted a variable that we didn't use.

yui-knk:
LGTM! Thank you!

PR #408, opened 2024-05-01
Add support for the callee side tag specification of parameterizing rules

ydah:
resolve: https://github.com/ruby/ruby/pull/10667#issuecomment-2087953669


yui-knk:
LGTM! Thank you!

PR #410, opened 2024-05-02
Add support for aliased Named References for actions of rhs in Parameterizing rules

ydah:
This PR support for aliased Named References for actions of rhs in Parameterizing rules.

## Motivation

If there is a non-terminal symbol in the RHS of parameterizing rules as shown below, I want to access it in the Named Reference.

```
%rule sum(X, Y): X[summand] '+' Y[addend] { $$ = $summand + $addend }
                ;
```

### DIFF
--- a/spec/fixtures/parameterizing_rules/user_defined/with_action_and_named_references.y
+++ b/spec/fixtures/parameterizing_rules/user_defined/with_action_and_named_references.y
@@ -16,7 +16,7 @@ static int yyerror(YYLTYPE *loc, const char *str);
 %token <i> number
 %token <s> string
 
-%rule pair(X, Y): X ',' Y { printf("(%d, %d)\n", $X, $3); }
+%rule pair(X, Y): X ',' Y[alias] { printf("(%d, %d)\n", $X, $alias); }
                 ;
 
 %%
--- a/spec/fixtures/parameterizing_rules/user_defined/with_action_and_named_references.y
+++ b/spec/fixtures/parameterizing_rules/user_defined/with_action_and_named_references.y
@@ -10,18 +10,17 @@ static int yyerror(YYLTYPE *loc, const char *str);
 
 %union {
     int i;
-    char *s;
 }
 
 %token <i> number
-%token <s> string
+%token <i> summand
 
-%rule pair(X, Y): X ',' Y[alias] { printf("(%d, %d)\n", $X, $alias); }
-                ;
+%rule plus(X): X '+' number[addend] { $$ = $X + $addend; }
+             ;
 
 %%
 
-program         : pair(number, string) { printf("pair odd even\n"); }
+program         : plus(summand) { printf("plus number\n"); }
                 ;
 
 %%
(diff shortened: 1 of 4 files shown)


yui-knk:
Can I ask the use case of this functionality? We can use arbitrary name for the variable (`X`) then if we want to access `Y` by `$alias`, we can write `%rule pair(X, alias):`.

ydah:
@yui-knk I changed it to an appropriate example in PR description. For example, I think it makes sense to set an alias for nterms, which is not a parameter in rhs for parameterization rules

spec/fixtures/parameterizing_rules/user_defined/with_action_and_named_references.y
```diff
@@ -16,7 +16,7 @@ static int yyerror(YYLTYPE *loc, const char *str);
 %token <i> number
 %token <s> string
 
-%rule pair(X, Y): X ',' Y { printf("(%d, %d)\n", $X, $3); }
+%rule pair(X, Y): X ',' Y[alias] { printf("(%d, %d)\n", $X, $alias); }
```
yui-knk, line 19: The [intention](https://github.com/ruby/lrama/pull/410#issuecomment-2093981195) makes sense for me. Could you update the test case grammar to match with the intention?

yui-knk:
IIUC aliased Named References for non parameter in RHS is already supported.

spec/fixtures/parameterizing_rules/user_defined/with_action_and_named_references.y
ydah: Indeed it is. I updated this PR.

ydah:
@yui-knk Is there any blocker left for this pull request?

yui-knk:
@ydah I'm bit confused because:

1) It seems the test case passes without source code changes

```diff
diff --git a/lib/lrama/grammar/binding.rb b/lib/lrama/grammar/binding.rb
index 21b6ad5..d215ad0 100644
--- a/lib/lrama/grammar/binding.rb
+++ b/lib/lrama/grammar/binding.rb
@@ -16,7 +16,7 @@ module Lrama
           resolved_args = symbol.args.map { |arg| resolve_symbol(arg) }
           Lrama::Lexer::Token::InstantiateRule.new(s_value: symbol.s_value, location: symbol.location, args: resolved_args, lhs_tag: symbol.lhs_tag)
         else
-          parameter_to_arg(symbol) || symbol
+          @parameter_to_arg[symbol.s_value] || symbol
         end
       end
```

```shell
$ bundle exec rake
...
Finished in 9.62 seconds (files took 0.27494 seconds to load)
219 examples, 0 failures
```

2) It seems PR description is misaligned

> If there is a non-terminal symbol in the RHS of parameterizing rules as shown below,

```
%rule plus_n(X): X term[number] { $$ = $1 + $number }
                ;
```

But I guess `term` might be terminal (!= non-terminal).

Could you check them?

ydah:
1. The test case was incorrect.
Correctly, specifying a Named Reference for a parameter results in the following error:

```
Failures:

  1) Lrama::Parser#parse when parameterizing rules when user defined rules with action with named references expands parameterizing rules
     Failure/Error: raise location.generate_error_message(message)

     RuntimeError:
       parameterizing_rules/user_defined/with_action_and_named_references.y:17:53: Referring symbol `summand` is not found.
       %rule sum(X, Y) <i>: X[summand] '+' Y[addend] { $$ = $summand + $addend; }
```

2. Fixed an incorrect test case, so fixed the example. What do you think?

yui-knk:
LGTM!

PR #412, opened 2024-05-02
Refactor for parameterizing_rule_resolver

ydah:


lib/lrama/grammar.rb
```diff
@@ -289,6 +293,18 @@ def append_special_symbols
       @accept_symbol = term
     end
 
+    def resolve_inline_rules
```
yui-knk, line 296: Is this change intentional? It seems that this method is not called from anywhere, at least `bundle exec rake` successes even so deleting this method.

ydah: sorry. I updated this PR.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
yui-knk, line 1: Could you remove this file from this commit?

yui-knk (approved):

.vscode/settings.json
ydah: 🙏

PR #414, opened 2024-05-04
Add support for Named Reference in parameterizing rules callers

ydah:
This PR add support for Named Reference in parameterizing rules callers.

## Motivation

In some cases I wanted to use alias when we defined compstmt as parameterizing rules:
https://github.com/ruby/ruby/blob/96710a3139fecda4bc12cd4f321399f71887535c/parse.y#L3263-L3275

```
| k_END allow_exits '{' compstmt '}' <--- when we want to replace compstmt with parameterizing rules
    {
        if (p->ctxt.in_def) {
            rb_warn0("END in method; use at_exit");
        }
        restore_block_exit(p, $allow_exits);
        p->ctxt = $k_END;
        {
            NODE *scope = NEW_SCOPE2(0 /* tbl */, 0 /* args */, $compstmt /* body */, &@$); <--- For example, it would be nice if parameterizing rules could be accessed as compstmt using alias, even if the name is not compstmt.
            $$ = NEW_POSTEXE(scope, &@$);
        }
    /*% ripper: END!($:compstmt) %*/
    }
```


ydah:
This change eliminates the following parsing error:
https://github.com/ruby/ruby/pull/10726/files#r1590905311

parser.y
```diff
@@ -444,17 +444,17 @@ rule
            builder.add_rhs(token)
            result = builder
          }
-     | rhs symbol parameterizing_suffix TAG?
+     | rhs symbol parameterizing_suffix named_ref_opt TAG?
```
yui-knk, line 447: Could you add test cases for this patten?

yui-knk:
LGTM, let's add test cases and rebase recent master branch.

yui-knk (approved):

parser.y
ydah: Thank you! Add test case: https://github.com/ruby/lrama/pull/414/commits/effe94adab3babaa1646d94ae4726cd5246a86fd

PR #418, opened 2024-05-09
[WIP] [PoC] Intoroduce parameterizing rules with conditonal

ydah:
I would like to propose a new grammar in this PR.
I believe that more parameterizing rules can handle more abstract rules if we can switch between rules and actions that are expanded by conditions in order to make rules common.

Syntax is as follows:
```
%rule defined_rule(X, condition): /* empty */
                                | X { $$ = $1; } %if(condition) /* 1 */
                                | %if(condition) X %endif X { $$ = $1; } /* 2 */
                                ;

%%

r_true        : defined_rule(number, %true)
              ;

r_false       : defined_rule(number, %false)
              ;
```

1. It's like a postfix if in Ruby. If condition is false, it is equivalent to missing this line.
2. If statementIf condition is false, it is equivalent to missing RHS between `%if` and`% endif`.

## Motivation

I believe it will solve the problem mentioned in the article below with the tight coupling with Lexer "to disable certain generation rules under certain conditions" and I would like to propose this feature to solve this problem.
https://yui-knk.hatenablog.com/entry/2023/04/04/190413

We can trace the RHS to [f_args](https://github.com/ruby/ruby/blob/2f916812a9b818b432ee7c299e021ec62d4727fb/parse.y#L5523-L5575) > [args_tail](https://github.com/ruby/ruby/blob/2f916812a9b818b432ee7c299e021ec62d4727fb/parse.y#L5487-L5503) > [args_forward](https://github.com/ruby/ruby/blob/2f916812a9b818b432ee7c299e021ec62d4727fb/parse.y#L5586-L5597), where f_args is the RHS of both the lambda argument (f_larglist) and the method definition argument (f_arglist).
So if we can switch between RHS and actions by passing parameters, we can break up the Lexer/Parser coupling here.


parser.y
```diff
@@ -253,14 +253,16 @@ rule
...
                 {
                   builder = val[0]
+                  builder.symbols << val[1] if val[1]
                   result = [builder]
                 }
-          | rule_rhs_list "|" rule_rhs
+          | rule_rhs_list "|" rule_rhs if_clause?
                 {
                   builder = val[2]
+                  builder.symbols << val[3] if val[3]
```
yui-knk, line 265: `builder` (is the instance of `Grammar::ParameterizingRule::Rhs`) has one or zero `if_clause`. Then I think it's reasonable for `builder` has attribute for `if_clause` than holding `if_clause` on `symbols`.

yui-knk:
I agree to the feature.
What I concern is the syntax. There are 3 points for discussion:

1. Modifier style (`%if(condition)`) makes senes for me
2. The syntax for parameters (`defined_rule(X, condition)`) is a point of concern. Parameters for rule (`X`) are different from parameters for condition (`condition`). If so I want to make the difference clear from the syntax. For example, in Ruby optional parameter is clear from its syntax (`a = 1`) and we can set the constrain on the order of parameters if condition parameter has different syntax from rule parameter. One example is adding some prefix to condition parameter like `defined_rule(X, @condition)`.
3. The syntax for condition argument is another point of concern. I think the meaning of rule argument and condition argument are different. Condition argument might be a kind of expression. If so it might be better to introduce different syntax like `defined_rule(number, %{true})` or `defined_rule(number, {true})` and so on.

Sorry for the late response.

PR #419, opened 2024-05-12
Use `require_relative` in the Lrama codebase

koic:
If there are many searches in the `$LOAD_PATH` in the user environment, require will perform unnecessary searches that are not needed. In contrast, `require_relative` is efficient because it uses a relative path.

Lrama requires Ruby 2.5+, it is possible to use `require_relative`, which was introduced in Ruby 1.9.
https://github.com/ruby/lrama/blob/v0.6.9/lrama.gemspec#L14


ydah (approved):
Thank you!

PR #420, opened 2024-05-17
Widen the definable position of parameterizing rules

ydah:
We came to this decision after discussion at RubyKaigi 2024. Definitions are also possible in the following positions:

```
%rule defined_option(X): /* empty */
                       | X
                       ;

%%

program         : defined_option(number) <i>
                | defined_list(number) <i>
                ;

%rule defined_list(X): /* empty */  /* <--- here */
                     | defined_list(X) number
                     ;
```


yui-knk:
I'm wondering which syntax is better for parameterizing rules definitions in `grammar` part, the part after `%%`.

## Option 1. With `%rule`

Current syntax then requires `%rule` even so in `grammar` part.

## Option 2. Without `%rule`

Like `defined_list(X): /* empty */  /* <--- here */ ...` in `grammar` part.

I prefer Option 2 because in `bison_declarations` part, it's not clear without `%rule` but in `grammar` part, it seems verbose to write `%rule`. Especially, if we treat normal rule as a "function without argument", it's natural for me to not require `%rule`.
What do you think about it?

ydah:
@yui-knk I also thought Option2 was the way to go, but there was one annoying problem with Option2.
[Since the `;` can be omitted in the case of generating rules](https://github.com/ruby/lrama/commit/2dd2aab2b0523bb113a0983db497b4edd83dd8e0), there is an SR conflict between the Parameterizing Rules at the end of the RHS and the Parameterizing Rules in the LHS.

The solution seems to be to either change the `;` in the generating rules to be non omittable or to make the `%rule` mandatory.

ydah:
On 6/25 (Wed) we discussed it face to face and decided that keeping compatibility was more important. `%rule` is now mandatory. cc/ @yui-knk @junk0612

yui-knk (approved):

PR #422, opened 2024-05-20
Use `foo?` instead of `foo_opt`

ydah:
## Motivation

racc's v1.8.0 has been released:
https://github.com/ruby/racc/releases/tag/v1.8.0

So we got a new grammar.
- https://github.com/ruby/racc/pull/222

This pull request uses `Option:? `. WDYT?

see: https://www.antlr.org/


yui-knk:
LGTM, this is what I expected!

PR #423, opened 2024-05-23
Use `string_as_id` instead of `STRING`

ydah:
The TODO added in the following PR will be eliminated: 
- https://github.com/ruby/lrama/pull/143


yui-knk:
Thank you!

PR #426, opened 2024-06-02
Rename Lrama::Grammar::Rule#to_s to Lrama::Grammar:Rule::display_name

S-H-GAMELINKS:
I saw a TODO in `lib/lrama/grammar/rule.rb` to rename the `to_s` method to the `display_name` method, so I renamed it.


ydah (approved):
Looks good! Thank you!

PR #427, opened 2024-06-02
Use `StringIO.new` instead of `""`

ydah:
Follow up: https://github.com/ruby/lrama/pull/424


junk0612 (approved):
It looks good to me. Thanks for following up.

PR #431, opened 2024-06-05
Ignore temporary parse files in .gitignore

ydah:
When measuring by stackprof, I always follow the procedure described in the README.
At that time, parse.tmp.c and parse.tmp.h are created, but I don't want to commit them by mistake.

So, ignore the file created by the following procedure: https://github.com/ruby/lrama/blob/24786d88a35b7f7e7b95963d18d20125b29eb949/README.md?plain=1#L159-L161

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 /Gemfile.lock
 /coverage/
 /parser.output
+/parse.tmp.*
 /pkg/
 /tmp/
 /vendor/bundle


junk0612:
I am not sure what this PR is trying to solve, so I cannot determine its merits. Can you explain what you see as the issue?

ydah:
I see. When measuring by stackprof, I always follow the procedure described in the README.
At that time, parse.tmp.c and parse.tmp.h are created, but I don't want to commit them by mistake.

junk0612:
I understand. Personally, I think this change might not be very effective since it could become meaningless if, for example, the file names are changed in an individual environment or if the README is updated. It seems more appropriate to address this in individual environments. However, I do not have strong objections, so if you still wish to include this change, I will not oppose it.

ydah:
@junk0612 I certainly heard your opinion and thought it better to have it set up as a personal setting. Thank you for your feedback! I will close this PR 👍🏻

PR #433, opened 2024-06-06
Add types in and around Grammar.rb

Little-Rubyist:
`Grammar.rb` has been added to Steep's monitoring targets. Along with this, we have defined many types.
Additionally, we have modified implementations where type consistency could not be maintained (the changes have been discussed with @yui-knk).
We have refactored parts where the return value could include nil so that Steep can correctly detect them.


lib/lrama/grammar/symbols/resolver.rb
```diff
@@ -44,7 +44,9 @@ def add_term(id:, alias_name: nil, tag: nil, token_id: nil, replace: false)
         end
 
         def add_nterm(id:, alias_name: nil, tag: nil)
-          return if find_symbol_by_id(id)
+          if (sym = find_symbol_by_id(id))
+            return sym
```
junk0612, line 48: I was wondering why you changed it to return `sym`, but I see it's because the previous version could return `nil`, which would fail type checking. Understood. It also aligns the behavior with `add_term`, which seems good.

junk0612 (approved):
Looks good! Thank you for your contribution to the developer experience.

PR #437, opened 2024-06-07
Fix some rbs

ydah:
- Fix an indentation in sig/lrama/options.rbs
- Fix rbs validate warning
  ```
  ❯ bundle exec rbs -I sig validate
  W, [2024-06-07T16:09:23.137310 #4825]  WARN -- rbs: sig/lrama/grammar.rbs:74:19...74:37: `void` type is only allowed in return type or generics parameter (RBS::WillSyntaxError)
        def validate!: () -> (void | bot)
                       ^^^^^^^^^^^^^^^^^^
  ```
- Use void instead of bot?

### DIFF
--- a/sig/lrama/grammar.rbs
+++ b/sig/lrama/grammar.rbs
@@ -71,7 +71,7 @@ module Lrama
     def epilogue_first_lineno=: (Integer epilogue_first_lineno) -> Integer
     def epilogue=: (String epilogue) -> String
     def prepare: () -> void
-    def validate!: () -> (void | bot)
+    def validate!: () -> void
     def find_rules_by_symbol!: (Grammar::Symbol sym) -> Array[Rule]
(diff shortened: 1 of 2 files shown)


sig/lrama/grammar.rbs
```diff
@@ -88,6 +88,6 @@ module Lrama
     def fill_default_precedence: () -> void
     def fill_symbols: () -> Array[Grammar::Symbol]
     def fill_sym_to_rules: () -> Array[Rule]
-    def validate_rule_lhs_is_nterm!: () -> bot?
+    def validate_rule_lhs_is_nterm!: () -> void
```
ydah, line 91: memo) The bot is used for methods that either always throw an exception or enter an infinite loop and never terminate. Therefore, I believe void is correct here.

PR #439, opened 2024-06-08
Add support `--report=terms` option

S-H-GAMELINKS:
Add report_unused_terms private method to Lrama::StatesReporter for `TODO Unused terms` in `lib/lrama/states_report.rb`
However, I am not sure if the output format is what it is supposed to be, so I would appreciate your comments if it is different.

### DIFF
--- a/lib/lrama/option_parser.rb
+++ b/lib/lrama/option_parser.rb
@@ -92,7 +92,7 @@ module Lrama
     end
 
     BISON_REPORTS = %w[states itemsets lookaheads solved counterexamples cex all none]
-    OTHER_REPORTS = %w[verbose]
+    OTHER_REPORTS = %w[terms verbose]
     NOT_SUPPORTED_REPORTS = %w[cex none]
     VALID_REPORTS = BISON_REPORTS + OTHER_REPORTS - NOT_SUPPORTED_REPORTS
 
--- a/lib/lrama/states_reporter.rb
+++ b/lib/lrama/states_reporter.rb
@@ -16,13 +16,51 @@ module Lrama
 
     private
 
-    def _report(io, grammar: false, states: false, itemsets: false, lookaheads: false, solved: false, counterexamples: false, verbose: false)
-      # TODO: Unused terms
+    def _report(io, grammar: false, terms: false, states: false, itemsets: false, lookaheads: false, solved: false, counterexamples: false, verbose: false)
       # TODO: Unused rules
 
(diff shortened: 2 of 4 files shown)


ydah (commented):
Thank you! This is a very useful reporting feature!
I think the implementation is good. However, isn't it not possible to execute it as a command since there is no `term` option?
https://github.com/ruby/lrama/blob/5bed7c91c4368fed853e8d1a8c19fe361131674d/lib/lrama/option_parser.rb#L94-L97

S-H-GAMELINKS:
Certainly, I had forgotten about the part about passing in terms option for command.
So I commited and force-pushed that implemented passing terms option for command.

lib/lrama/option_parser.rb
```diff
@@ -91,7 +91,7 @@ def parse_by_option_parser(argv)
       end
     end
 
-    BISON_REPORTS = %w[states itemsets lookaheads solved counterexamples cex all none]
+    BISON_REPORTS = %w[terms states itemsets lookaheads solved counterexamples cex all none]
     OTHER_REPORTS = %w[verbose]
```
ydah, line 95: Could you move to `OTHER_REPORTS`?

It because below:

```
❯ bison -h
Usage: bison [OPTION]... FILE
: (snip)
Output Files:
  -H, --header=[FILE]           also produce a header file
  -d                            likewise but cannot specify FILE (for POSIX Yacc)
  -r, --report=THINGS           also produce details on the automaton
: (snip)
THINGS is a list of comma separated words that can include:
  states            describe the states
  itemsets          complete the core item sets with their closure
  lookaheads        explicitly associate lookahead tokens to items
  solved            describe shift/reduce conflicts solving
  counterexamples, cex
                    generate conflict counterexamples
  all               include all the above information
  none              disable the report
```

```suggestion
    BISON_REPORTS = %w[terms states itemsets lookaheads solved counterexamples cex all none]
    OTHER_REPORTS = %w[verbose]
```

S-H-GAMELINKS: OK, it's moved.

ydah (approved):
Looks fine! Thank you!

ydah:
@S-H-GAMELINKS Ah, sorry. Could you change the title of the commit message and PR to say that you have added support for the `--report=terms` option?

S-H-GAMELINKS:
OK, I have rewritten PR titile and commit message.

ydah (approved):
Thank you!

PR #448, opened 2024-06-18
Introduce the `-W` and `--warnings` options and support redefined parameterizing rules

ydah:
This PR adds an option to output warnings and adds an option to warn when parameterizing rules are being redefined.
Specify `-W` or `--warning` option to output a warning about the redefinition of parameterizing rules.

```
Diagnostics:
    -W, --warnings         report the warnings
```

## Motivation

Currently, for example, there is no way for users to know that the parameterizing rules provided by the standard library have been redefined.
Therefore, this support will allow users to know when parameterizing rules have been redefined.


yui-knk:
I talked with @ydah in person on 6/25 (Wed). I agreed to add `-W` option and `--warnings` option, but I thought introducing the categories are premature because we are not sure about what categories are the best. We will introduce warning categories once we get feedback for that.

PR #458, opened 2024-07-16
Add types to classes in state directory

Little-Rubyist:


sig/lrama/states/item.rbs
```diff
@@ -0,0 +1,44 @@
+module Lrama
+  class States
+    class Item
```
Little-Rubyist, line 3: It is not subject to Item class type checking. I'll define the type in another PR.

lib/lrama/state/resolved_conflict.rb
```diff
@@ -8,7 +8,7 @@ class State
     class ResolvedConflict < Struct.new(:symbol, :reduce, :which, :same_prec, keyword_init: true)
       def report_message
         s = symbol.display_name
-        r = reduce.rule.precedence_sym.display_name
+        r = reduce.rule.precedence_sym&.display_name
```
Little-Rubyist, line 11: It seemed to be nullable. is it acceptable to have nil in the `r` variable?
Currently allows nil to pass type checking.

sig/lrama/states/item.rbs
ydah: imo) If it's TODO, I think it's better to add a TODO comment because if for some reason you can't send a follow up PR, someone can notice this later. But, it is not a strong opinion.

lib/lrama/state/reduce.rb
```diff
@@ -27,6 +27,7 @@ def add_not_selected_symbol(sym)
 
       def selected_look_ahead
         if @look_ahead
+          # @type ivar @look_ahead: Array<Grammar::Symbol>
```
ydah, line 30: ask) Is this an intended change?

sig/lrama/states/item.rbs
Little-Rubyist: Well, sure. Thanks for the opinion.
However, I can get a PR out immediately when this case. I will write a TODO next time.

lib/lrama/state/reduce.rb
ydah: Thanks for the explanation! Understood.

ydah (approved):
Thank you

PR #463, opened 2024-07-30
Add types to states

Little-Rubyist:
This is the type definition of Item.rb added in [this PR](https://github.com/ruby/lrama/pull/458#discussion_r1682331387).

### DIFF
--- a/lib/lrama/states/item.rb
+++ b/lib/lrama/states/item.rb
@@ -56,11 +56,11 @@ module Lrama
         Item.new(rule: rule, position: position + 1)
       end
 
-      def symbols_before_dot
+      def symbols_before_dot # steep:ignore
         rhs[0...position]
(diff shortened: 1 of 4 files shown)


lib/lrama/states/item.rb
```diff
@@ -56,11 +56,11 @@ def new_by_next_position
         Item.new(rule: rule, position: position + 1)
       end
 
-      def symbols_before_dot
+      def symbols_before_dot # steep:ignore
```
Little-Rubyist, line 59: `Array#[]` is nullable but `symbols_before_dot` and `symbols_after_dot` are not.
But I couldn't make `Array#[]` temporarily not null, so I added `steep:ignore` .

ydah (approved):
Thank you!

Issue #442, opened 2024-06-10
Error in initialization within YY_INITIAL_VALUE() with C++


matz:
In the generated C file, we see `YY_INITIAL_VALUE (static YYSTYPE yyval_default;)`, but `clang` raises error `error: default initialization of an object of const type`. According to the C++ language specification, we need a user provided default constructor for this kind of declaration.

 Changing the above line to 'YY_INITIAL_VALUE (static YYSTYPE yyval_default;)' as Bison 3.8.2 shows no error.

matz:
Unfortunately

```
static const YYSTYPE yyval_default = YY_INITIAL_VALUE(YYSTYPE());
```

causes compilation error when `#define YY_INITIAL_VALUE(Value) /* Nothing. */`

```
error: expected primary-expression before ‘;’ token
 6396 | static const YYSTYPE yyval_default = YY_INITIAL_VALUE(YYSTYPE());
      |                                                                 ^
```

PR #447, opened 2024-06-18
Add CI env for C++

yui-knk:
This is a follow up of https://github.com/ruby/lrama/pull/443

PR #465, opened 2024-08-09
Add CI env for C++ and fix errors and warnings when compiling as C++

ydah:
follow up: https://github.com/ruby/lrama/pull/447
fix: https://github.com/ruby/lrama/issues/442


yui-knk:
LGTM! Thank you!

PR #475, opened 2024-10-01
Use grouping instead of generic_symbol

ydah:


yui-knk (approved):

spec/fixtures/context/basic.y
```diff
@@ -18,6 +18,13 @@
 %printer {
     print_token();
 } tNUMBER tSTRING
+%destructor {
```
yui-knk, line 21: Is this intentional change or forget to remove after debug?

ydah: Intentionally change it. I noticed in spec/lrama/context_spec.rb that `.flatten` is needed in `%printer`. So I also added `%destructor` and `%error-token`. However, I am sure that they are not essential tests in this file, so you may want to check them with other tests. What do you think?

PR #478, opened 2024-11-07
Add warning output when using %require, which exists only for compatibility with Bison

ydah:
This PR adds a warning output when using %require, which exists only for compatibility with Bison. 
see: https://www.gnu.org/software/bison/manual/html_node/Require-Decl.html

### DIFF
--- a/spec/lrama/warnings/required_spec.rb
+++ b/spec/lrama/warnings/required_spec.rb
@@ -1,0 +1,49 @@
+# frozen_string_literal: true
+
+RSpec.describe Lrama::Warnings::Required do
+  describe "#warn" do
+    let(:y) do
+      <<~STR
+        %require "3.0"
+        %{
+        // Prologue
+        %}
+        %union {
+            int i;
+        }
+        %token <i> tNUMBER
+        %%
+        program: tNUMBER
+                ;
+      STR
+    end
+
+    context "when warnings true" do
+      it "has warns for require is provided for compatibility with bison" do
+        grammar = Lrama::Parser.new(y, "states/required.y").parse
+        grammar.prepare
+        grammar.validate!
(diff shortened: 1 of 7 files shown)


yui-knk:
Right now, I don't have any concrete direction about `%require`, implement it or deprecate it.

ydah:
@yui-knk Thank you feedback. If I had one concern, it would be that when considering Lrama not as a parser generator for CRuby but as a Bison-compatible parser generator, users might mistakenly believe they can control it with `%require`. This could lead to misunderstandings, which was my motivation for bringing it up.  

For example, how about the following message?  

*"Currently, `%require` is simply valid as a grammar but does nothing."*

yui-knk:
The new message seems fine to me because it only explains current status.

ydah:
@yui-knk I updated  message. Thank you.

spec/lrama/warnings/required_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+      it "has not warns for require is provided for compatibility with bison" do
+        grammar = Lrama::Parser.new(y, "states/required.y").parse
+        grammar.prepare
+        grammar.validate!
+        states = Lrama::States.new(grammar, Lrama::Tracer.new(Lrama::Logger.new))
+        states.compute
+        logger = Lrama::Logger.new
+        allow(logger).to receive(:warn)
+        Lrama::Warnings.new(logger, false).warn(grammar, states)
+        expect(logger).not_to have_received(:warn).with("currently, %require is simply valid as a grammar but does nothing")
```
yui-knk, line 45: It's better to check `logger#warn` is not called regardless its message otherwise the test case will pass when different message is send.

yui-knk (approved):

spec/lrama/warnings/required_spec.rb
ydah: Indeed. Thank you!

PR #483, opened 2024-12-13
Add types to counterexamples

Little-Rubyist:


Little-Rubyist:
@ydah Thank you for pointing this out!
I have resolved the conflict, please review again :bow:

sig/lrama/states.rbs
```diff
@@ -0,0 +1,143 @@
+module Lrama
+  # States is passed to a template file
```
yui-knk, line 2: It seems these comments come from "lib/lrama/states.rb". Are they needed for rbs file?

lib/lrama/counterexamples/path.rb
```diff
@@ -20,6 +20,10 @@ def to_s
         "#<Path(#{type})>"
       end
       alias :inspect :to_s
+
+      def type
+        raise NoMethodError
```
yui-knk, line 25: It's expected for sub classes to implement `#type` then `NotImplementedError` is better than `NoMethodError`.

lib/lrama/counterexamples/example.rb
```diff
@@ -57,12 +58,14 @@ def _derivations(paths)
             when ProductionPath
               derivation = Derivation.new(item, derivation)
               current = :production
+            else
+              return nil
```
yui-knk, line 62: If I remember correctly, `path` should be one of StartPath, TransitionPath or ProductionPath then it's better to `raise "Unexpected. #{path}"` than return nil.

We can omit `?` for some methods by this change.

```diff
diff --git a/lib/lrama/counterexamples/example.rb b/lib/lrama/counterexamples/example.rb
index 01995ab..bb08428 100644
--- a/lib/lrama/counterexamples/example.rb
+++ b/lib/lrama/counterexamples/example.rb
@@ -38,7 +38,7 @@ module Lrama
       private

       def _derivations(paths)
-        derivation = nil #: Derivation?
+        derivation = nil #: Derivation
         current = :production
         last_path = paths.last #: Path
         lookahead_sym = last_path.to.item.end_of_rule? ? @conflict_symbol : nil
@@ -59,7 +59,7 @@ module Lrama
               derivation = Derivation.new(item, derivation)
               current = :production
             else
-              return nil
+              raise "Unexpected. #{path}"
             end

             if lookahead_sym && item.next_next_sym && item.next_next_sym.first_set.include?(lookahead_sym)
diff --git a/sig/lrama/counterexamples/example.rbs b/sig/lrama/counterexamples/example.rbs
index ba55d46..fae1a98 100644
--- a/sig/lrama/counterexamples/example.rbs
+++ b/sig/lrama/counterexamples/example.rbs
@@ -31,13 +31,13 @@ module Lrama

       def path2_item: () -> States::Item

-      def derivations1: () -> Derivation?
+      def derivations1: () -> Derivation

-      def derivations2: () -> Derivation?
+      def derivations2: () -> Derivation

       private

-      def _derivations: (::Array[StartPath | TransitionPath | ProductionPath] paths) -> Derivation?
+      def _derivations: (::Array[StartPath | TransitionPath | ProductionPath] paths) -> Derivation

       def find_derivation_for_symbol: (StateItem state_item, Grammar::Symbol sym) -> Derivation?
     end
```

sig/lrama/states.rbs
Little-Rubyist: Comments were included because they were generated by rbs prototype and committed without change.
I have deleted comments that seem unnecessary.

lib/lrama/counterexamples/example.rb
Little-Rubyist: I have set you as co-author and committed. Thanks for pointing that out.

yui-knk:
Thank you!

PR #484, opened 2024-12-18
Refactoring with option in parser.y

ydah:
Use new racc's grammar: `?`

### DIFF
--- a/spec/lrama/diagnostics_spec.rb
+++ b/spec/lrama/diagnostics_spec.rb
@@ -30,7 +30,7 @@ RSpec.describe Lrama::Diagnostics do
 
           %type <i> class /* comment for class */
           %nterm <str> string /* comment for string */
-          %type string_1 string_2 /* missing `%type` tag */
+          %type string_1 string_2 /* <tag> is omitted */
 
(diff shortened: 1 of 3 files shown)


spec/lrama/diagnostics_spec.rb
```diff
@@ -29,6 +29,7 @@
           %token tEQEQ  "=="
 
           %type <i> class /* comment for class */
+          %type string_1 string_2 /* missing `%type` tag */
```
yui-knk, line 32: What this comment means? I think `%type` is not missed on this grammar file.

ydah: What I meant to say there is that the `%type` tag is missing. In other words, I wanted to make it clear that I was testing the lack of `<tag>`.

yui-knk: I got it. What do you think about comment like `/* <tag> is omitted */` ?

ydah: Thank you! I updated comment.

yui-knk (approved):

ydah:
resolve conflict

PR #490, opened 2024-12-25
Use rbs-inline and rewrite bitmap.rb

Little-Rubyist:
I'll be installing rbs-inline, which we discussed previously.
rbs-inline requires writing the type information to `.rb` files and generating `.rbs` files with `rbs-inline --output`.
ref: https://github.com/soutaro/rbs-inline?tab=readme-ov-file#usage

I were also considering installing rubocop-on-rbs, but decided to pass on it because rubocop has not use yet.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -16,5 +16,6 @@ gem "memory_profiler"
 # Then skip install on some CI jobs.
 if !ENV['GITHUB_ACTION'] || ENV['INSTALL_STEEP'] == 'true'
   gem "rbs", "3.7.0", require: false
+  gem "rbs-inline", require: false
   gem "steep", "1.9.1", require: false
 end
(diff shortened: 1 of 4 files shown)


yui-knk:
Just in case, I want to confirm the development process.
Will we generate rbs file by `rbs-inline --output` on local and commit generated files so that released gem will include rbs files same as before?

Little-Rubyist:
> Just in case, I want to confirm the development process. Will we generate rbs file by `rbs-inline --output` on local and commit generated files so that released gem will include rbs files same as before?

@yui-knk 
Yes, that's correct.
If we forget to generate, we should be able to notice because the `steep check` will fail.

yui-knk:
I got it, thank you!

PR #501, opened 2025-01-02
Move processing of lhs_s_value method to Binding class

ydah:
This PR move processing of lhs_s_value method to Binding class.

### DIFF
--- a/lib/lrama/grammar/binding.rb
+++ b/lib/lrama/grammar/binding.rb
@@ -22,6 +22,10 @@ module Lrama
         end
       end
 
+      def concatenated_args_str(token)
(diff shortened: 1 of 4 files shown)


lib/lrama/grammar/binding.rb
```diff
@@ -30,6 +34,17 @@ def parameter_to_arg(symbol)
         end
         arg
       end
+
+      def token_to_args_s_values(token)
+        token.args.flat_map do |arg|
```
yui-knk, line 39: Note: `Array#join` expands array of array recursively then both `#flat_map` and `#map` works. However returning unnested array is simpler.

yui-knk (approved):

PR #521, opened 2025-01-22
Add comments for State methods

yui-knk:


lib/lrama/state.rb
```diff
@@ -364,6 +381,10 @@ def goto_follow_set(nterm_token)
         .reduce(always_follows(shift, next_state)) {|result, terms| result |= terms }
     end
 
+    # Definition 3.14 (goto_follows, via successor_follows)
+    #
+    # TODO: This method uses `#always_follows` but the original papar uses `successor_follows`,
```
yui-knk, line 386: @junk0612 JFYI.

junk0612: @yui-knk 
What you say in your comment is true, but I used definition 3.24(goto_follows, via always_follows) for the implementation.
These sets are equivalent, as proved in Theorem 3.25.

yui-knk: I got it, thank you. Fixed by https://github.com/ruby/lrama/pull/522.

PR #523, opened 2025-02-01
Add diagram generation feature and integrate Railroad Diagrams

ydah:
## Overview

Lrama provides an API for generating HTML syntax diagrams. These visual diagrams are highly useful as grammar development tools and can also serve as a form of automatic self-documentation.

Syntax Diagrams

This feature was inspired by the functionality of [Chevrotain](https://chevrotain.io/docs/), a parser construction toolkit for JavaScript.

## Motivation
The main motivation for adding this feature was to make the structure of `parse.y`, and thus the structure of Ruby's grammar, easier to understand. I thought that generating syntax diagrams in a more human-readable format would be beneficial.

These diagrams are highly useful as grammar development tools and can also serve as automatic self-documentation, which I believe is a significant advancement.

## Library for Rendering Railroad Diagrams
Chevrotain used [tabatkins/railroad-diagrams](https://github.com/tabatkins/railroad-diagrams), but this library only supported JavaScript and Python.

To address this, I created a Ruby library and integrated it into Lrama: [https://github.com/ydah/railroad_diagrams](https://github.com/ydah/railroad_diagrams)

This makes collaboration with Lrama easier and allows improvements to the library based on Lrama's needs. For these reasons, I decided to release it as a Ruby gem.


lib/lrama/diagram.rb
```diff
@@ -0,0 +1,77 @@
+# frozen_string_literal: true
+
+require "erb"
+begin
+  require "railroad_diagrams"
+rescue LoadError
+  warn "railroad_diagrams is not installed. Please run `bundle install`."
+end
```
ydah, line 8: If you're using it in `ruby/ruby`, the library may not be installed by default. In such cases, you need to handle it like this:  

This ensures that the script doesn't break if the gem isn't available, while also notifying the user about the missing dependency.

But if there is a better way, I would like to know.

.github/workflows/test.yaml
```diff
@@ -26,7 +26,6 @@ jobs:
       - uses: ruby/setup-ruby@v1
         with:
           ruby-version: ${{ matrix.ruby }}
-          bundler-cache: true
```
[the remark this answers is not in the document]
yui-knk: Could comment out the line instead of deletion and add the link to failed CI job as comments?

NEWS.md
```diff
@@ -1,5 +1,13 @@
 # NEWS for Lrama
 
+## Lrama 0.7.1 (2025-xx-xx)
+
+### Syntax Diagrams
```
yui-knk, line 5: It might good to add how to specify the option.

lib/lrama/diagram.rb
yui-knk: Yes, we should not enforce any installation of gems for ruby/ruby build.
However current implementation causes warnings on ruby/ruby build (see:  https://github.com/ruby/lrama/actions/runs/13088787136/job/36523059170?pr=523#step:14:76), we also need to avoid such situation.
When I implemented [stackprof integration](https://github.com/ruby/lrama/blob/88097af10d937c5c6a2890a9be75a5b27f0d1c8d/lib/lrama/report/profile.rb#L8), I postponed `require` until entry point method is called.

Then my suggestions are:
1. extract `require` logic (L. 4 - 8) to an instance method
2. call the instance method when it's necessary, it might be needed when `Diagram#initialize` is called

By the way, we can assume "erb" is bundled to ruby then please keep `require "erb"` as is.

yui-knk: https://github.com/ruby/lrama/blob/1df1c1b5b3a4ea265415d134c9bfb7991887ae39/lib/lrama/report/profile/call_stack.rb#L24 is an example.

ydah: Thank you! I updated https://github.com/ruby/lrama/pull/523/commits/9113e0b873664f7eb1306ded809474e418251967

.github/workflows/test.yaml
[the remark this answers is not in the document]
ydah: I add comment: https://github.com/ruby/lrama/pull/523/commits/ffd311599a114ad69f12f2d619978facbda20a5d

NEWS.md
ydah: Indeed it is.  I updated: https://github.com/ruby/lrama/pull/523/commits/4280ca2aac11b7173a94ba95e03d827483ba2bd8

yui-knk (approved):

PR #527, opened 2025-02-04
Fix an error when specify `--define=foo`

ydah:
```
❯ exe/lrama --define=foo sample/calc.y
wrong array length at 0 (expected 2, was 1)
```

This is occurring and has been addressed below:
https://github.com/ruby/lrama/blob/7fa74d84a68b3632950b969d29ccf012a0ea4a42/lib/lrama/grammar.rb#L60

### DIFF
--- a/lib/lrama/options.rb
+++ b/lib/lrama/options.rb
@@ -6,11 +6,11 @@ module Lrama
   class Options
     attr_accessor :skeleton #: String
     attr_accessor :header #: bool
-    attr_accessor :header_file #: String
-    attr_accessor :report_file #: String
+    attr_accessor :header_file #: String?
+    attr_accessor :report_file #: String?
     attr_accessor :outfile #: String
     attr_accessor :error_recovery #: bool
-    attr_accessor :grammar_file #: String
+    attr_accessor :grammar_file #: String?
     attr_accessor :trace_opts #: Hash[Symbol, bool]?
     attr_accessor :report_opts #: Hash[Symbol, bool]?
     attr_accessor :diagnostic #: bool
--- a/lib/lrama/options.rb
+++ b/lib/lrama/options.rb
@@ -12,7 +12,7 @@ module Lrama
 
     def initialize
       @skeleton = "bison/yacc.c"
-      @define = []
+      @define = {}
       @header = false
       @header_file = nil
       @report_file = nil
--- a/lib/lrama/options.rb
+++ b/lib/lrama/options.rb
@@ -1,15 +1,27 @@
+# rbs_inline: enabled
 # frozen_string_literal: true
 
 module Lrama
   # Command line options.
   class Options
-    attr_accessor :skeleton, :header, :header_file,
-                  :report_file, :outfile,
-                  :error_recovery, :grammar_file,
-                  :trace_opts, :report_opts,
-                  :diagnostic, :y, :debug, :define,
-                  :diagram, :diagram_file, :profile_opts
+    attr_accessor :skeleton #: String
+    attr_accessor :header #: bool
+    attr_accessor :header_file #: String
+    attr_accessor :report_file #: String
+    attr_accessor :outfile #: String
+    attr_accessor :error_recovery #: bool
+    attr_accessor :grammar_file #: String
+    attr_accessor :trace_opts #: Hash[Symbol, bool]?
+    attr_accessor :report_opts #: Hash[Symbol, bool]?
+    attr_accessor :diagnostic #: bool
+    attr_accessor :y #: IO
+    attr_accessor :debug #: bool
+    attr_accessor :define #: Hash[String, String]
+    attr_accessor :diagram #: bool
+    attr_accessor :diagram_file #: String
+    attr_accessor :profile_opts #: Hash[Symbol, bool]?
 
+    # @rbs () -> void
     def initialize
       @skeleton = "bison/yacc.c"
       @define = {}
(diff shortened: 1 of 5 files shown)


yui-knk:
I'm not sure why steep didn't warn however please update https://github.com/ruby/lrama/blob/7fa74d84a68b3632950b969d29ccf012a0ea4a42/sig/lrama/options.rbs#L4 too.

yui-knk:
[Note] It seems steep doesn't warn on array assignment to Hash type instance variable:

```ruby
# code
@define = []

# rbs
attr_accessor define: Hash[String, String]
```

```shell
$ bundle exec steep check
# Type checking files:

......................................................................................................................

No type error detected. 🫖
```

but it warns on hash assignment to Array type instance variable:

```ruby
# code
@define = {}

# rbs
attr_accessor define: Array[String]
```

```shell
$ bundle exec steep check
# Type checking files:

.....................................................................F................................................

lib/lrama/options.rb:15:16: [warning] Empty hash doesn't have type annotation
│ Diagnostic ID: Ruby::UnannotatedEmptyCollection
│
└       @define = {}
                  ~~

Detected 1 problem from 1 file
```

junk0612:
This change alone is probably not sufficient, and changes are needed, for example, in the location of the assignment at below. https://github.com/ruby/lrama/blob/7fa74d84a68b3632950b969d29ccf012a0ea4a42/parser.y#L28
By the way, I think Hash is appropriate for define because it is written in the grammar file as `%define lr.type ielr` or from the command line as `lrama -D lr.type=ielr`, but why change it to Array?

ydah:
> This change alone is probably not sufficient, and changes are needed, for example, in the location of the assignment at below.

Changes there will be made separately. Certainly as it is now, if a value is missing, it will result in an error.

> By the way, I think Hash is appropriate for define because it is written in the grammar file as `%define lr.type ielr` or from the command line as `lrama -D lr.type=ielr`, but why change it to Array?

I didn't mean to change the `@define` to Array in the first place. Because, below, I am converting from Array to Hash.
https://github.com/ruby/lrama/blob/7fa74d84a68b3632950b969d29ccf012a0ea4a42/lib/lrama/grammar.rb#L60

This PR simply changes the default value from Hash to Array.
Because the following Option is received in Array.
https://github.com/ruby/lrama/blob/968725eec15d1c0967d6bc75490febf78d66fcae/lib/lrama/option_parser.rb#L63

If I understand something incorrectly, please point it out to me.

ydah:
I changed it back to Hash to avoid the error https://github.com/ruby/lrama/pull/527/commits/569f568853e051db866f929ccf83fa8b71d7b290
And moved to rbs-inline style type definition https://github.com/ruby/lrama/pull/527/commits/4bd35c084b9aa1324ff50097349e77473195d1ac

lib/lrama/options.rb
```diff
@@ -1,15 +1,27 @@
...
-    attr_accessor :skeleton, :header, :header_file,
-                  :report_file, :outfile,
-                  :error_recovery, :grammar_file,
-                  :trace_opts, :report_opts,
-                  :diagnostic, :y, :debug, :define,
-                  :diagram, :diagram_file, :profile_opts
+    attr_accessor :skeleton #: String
+    attr_accessor :header #: bool
+    attr_accessor :header_file #: String
+    attr_accessor :report_file #: String
```
yui-knk, line 10: It's initialized with `nil` then `String?`.

lib/lrama/options.rb
```diff
@@ -1,15 +1,27 @@
...
+    attr_accessor :outfile #: String
+    attr_accessor :error_recovery #: bool
+    attr_accessor :grammar_file #: String
```
yui-knk, line 13: It's initialized with `nil` then `String?`.

spec/lrama/option_parser_spec.rb
```diff
@@ -343,4 +343,24 @@
       end
     end
   end
+
+  describe "@define" do
```
yui-knk, line 347: Could you add a test case for the issue which this PR will resolve, in short test case for `--define=foo`?

junk0612:
> I didn't mean to change the @define to Array in the first place. Because, below, I am converting from Array to Hash. It looks like the intended code.

I see, indeed. My understanding was wrong.

lib/lrama/options.rb
```diff
@@ -1,15 +1,27 @@
...
   class Options
-    attr_accessor :skeleton, :header, :header_file,
-                  :report_file, :outfile,
-                  :error_recovery, :grammar_file,
-                  :trace_opts, :report_opts,
-                  :diagnostic, :y, :debug, :define,
-                  :diagram, :diagram_file, :profile_opts
+    attr_accessor :skeleton #: String
+    attr_accessor :header #: bool
+    attr_accessor :header_file #: String
```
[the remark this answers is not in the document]
ydah: Fixed: https://github.com/ruby/lrama/pull/527/commits/7e064dea26300767aa3d30e7fb49a72985c8a55b

lib/lrama/options.rb
ydah: Fixed: https://github.com/ruby/lrama/pull/527/commits/7e064dea26300767aa3d30e7fb49a72985c8a55b

lib/lrama/options.rb
ydah: Fixed: https://github.com/ruby/lrama/pull/527/commits/7e064dea26300767aa3d30e7fb49a72985c8a55b

spec/lrama/option_parser_spec.rb
ydah: Added: https://github.com/ruby/lrama/pull/527/commits/5dfb6a3f77e7877a3dc962bcb05efd6c50889b11

yui-knk (approved):

PR #534, opened 2025-02-09
Organized directories and classes

ydah:
The following were the main changes taken
- Organized classes related to report, trace, and warnings
- Organized `Lrama::Command`
- The parts of report and trace that used to use puts are now using `Lrama::Logger`.
- Several test enhancements and additional type definitions.


yui-knk:
I will take a look at this PR tomorrow.

lib/lrama/command.rb
```diff
@@ -5,65 +5,104 @@ class Command
...
-    def run(argv)
-      parser_options(argv)
+    def initialize(argv)
+      @logger = Lrama::Logger.new
+      @options = OptionParser.parse(argv)
+      @tracer = Tracer.new(@logger, **@options.trace_opts)
+      @reporter = Reporter.new(**@options.report_opts)
+      @warnings = Warnings.new(@logger, @options.warnings)
+    rescue => e
+      message = e.message
```
yui-knk, line 15: Can we use `#format_error_message`?

lib/lrama/command.rb
```diff
@@ -5,65 +5,104 @@ class Command
...
+
+      grammar.insert_before_parameterizing_rules(stdlib_grammar.parameterizing_rules)
+    end
+
+    def prepare_grammar(grammar)
+      grammar.prepare
+      grammar.validate!
+    end
+
+    def compute(grammar)
```
yui-knk, line 84: I prefer `#compute_states`.

lib/lrama/trace.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
```
yui-knk, line 1: What is the reason why keeping `Trace` module even so introducing `Tracer`?

lib/lrama/report.rb
```diff
@@ -1,4 +1,8 @@
 # frozen_string_literal: true
 
-require_relative 'report/duration'
+require_relative 'report/conflicts'
```
yui-knk, line 3: What is the reason why keeping `Report` module even so introducing `Reporter`?

lib/lrama/reporter.rb
```diff
@@ -0,0 +1,30 @@
...
+      @terms = Lrama::Report::Terms.new(**options)
+      @conflicts = Lrama::Report::Conflicts.new
+      @grammar = Lrama::Report::Grammar.new(**options)
+      @states = Lrama::Report::States.new(**options)
+    end
+
+    # @rbs (Lrama::States states, File io) -> void
+    def report(states, io)
+      report_duration(:report) do
+        logger = Lrama::Logger.new(io)
```
yui-knk, line 21: Logger is an abstraction for managing which level of information is written to `io`. However `Reporter` is expected to always write the information. Please don't wrap `io` for report.

lib/lrama/command.rb
```diff
@@ -5,65 +5,104 @@ class Command
     LRAMA_LIB = File.realpath(File.join(File.dirname(__FILE__)))
     STDLIB_FILE_PATH = File.join(LRAMA_LIB, 'grammar', 'stdlib.y')
 
-    def run(argv)
-      parser_options(argv)
+    def initialize(argv)
+      @logger = Lrama::Logger.new
+      @options = OptionParser.parse(argv)
+      @tracer = Tracer.new(@logger, **@options.trace_opts)
+      @reporter = Reporter.new(**@options.report_opts)
```
yui-knk, line 12: I prefer to pass `io` when `Reporter` is initialized.

lib/lrama/command.rb
```diff
@@ -5,65 +5,104 @@ class Command
     LRAMA_LIB = File.realpath(File.join(File.dirname(__FILE__)))
     STDLIB_FILE_PATH = File.join(LRAMA_LIB, 'grammar', 'stdlib.y')
 
-    def run(argv)
-      parser_options(argv)
+    def initialize(argv)
+      @logger = Lrama::Logger.new
+      @options = OptionParser.parse(argv)
+      @tracer = Tracer.new(@logger, **@options.trace_opts)
```
yui-knk, line 11: Could you keep to pass `io` without `Logger` wrapper? I think it's difficult to determine what severity is best for trace because trace is enabled by separated option.

lib/lrama/command.rb
```diff
@@ -5,65 +5,104 @@ class Command
...
-      message = message.gsub(/.+/, "\e[1m\\&\e[m") if Exception.to_tty?
-      abort message
+      raise e if @options.debug
+      abort format_error_message(e.message)
     end
 
-    def _run
-      Report::Duration.enable if @options.trace_opts[:time]
+    def format_error_message(message)
+      return message unless $stderr.tty?
```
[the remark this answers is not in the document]
ydah: Fix: b9016f371c43c63e18c03eb35eb02595225b6183

lib/lrama/command.rb
ydah: Yes, thank you. Change: ead9c9481b353365abcc3cff8b36780c6bc70e47

lib/lrama/command.rb
ydah: Fix in 36a2638d78e68f2e2cdbb9db9f43857bd9317e78

lib/lrama/command.rb
ydah: Fix in 76e59b9d30cadbcba51ebffc0973010b54b5bdd5

lib/lrama/reporter.rb
ydah: Fix in 62ca9028be619e6b3773cd95dbbf823f3c10f38b

lib/lrama/trace.rb
ydah: The discomfort of the call made it so. For example, the `Trace` class would be `Trace.trace` or `Trace.run`, but `Tracer.trace` or `Tracer.run` seemed more natural. What do you think?

lib/lrama/report.rb
ydah: Same as: https://github.com/ruby/lrama/pull/534/files#r1952220404

lib/lrama/command.rb
ydah: If I pass the result of File.open here and initialize it, it will be difficult to manage when I let it abort in the middle of the process, so I'm doing it the way it is now, but would it be better to initialize it here?

lib/lrama/trace.rb
yui-knk: My concern is that classes under `class Trace` are used only by `Tracer` then it's easy to understand if these classes are placed within `Tracer`.

lib/lrama/command.rb
yui-knk: Understand your point.

lib/lrama/report.rb
ydah: fixed in: https://github.com/ruby/lrama/pull/534/commits/62c5593fa2e83b2420dfd88259191b622f214ffa

ydah:
@yui-knk Thank you for your review. I updated this PR.

lib/lrama/trace.rb
ydah: refixed: https://github.com/ruby/lrama/pull/534/commits/4026c1dc1451f9d7ac7bf03838259d8e7edd4d0e

lib/lrama/report.rb
ydah: refixed: https://github.com/ruby/lrama/pull/534/commits/4026c1dc1451f9d7ac7bf03838259d8e7edd4d0e

yui-knk (approved):

PR #536, opened 2025-02-10
Refactor related parameterized rule and inline rule

ydah:
The following refactoring was performed
- Use "Parameterized" instead of "Parameterizing".
  - Reduce the number of feature name types and organize them a bit
- Implement inline rule resolution with dedicated resolver class.


ydah:
Resolve conflicts

yui-knk:
I will review this PR.

yui-knk:
I agree to use "parameterized" in code base, actually "parameterized" is used in menhir code base.
One point need consideration is the name of this feature https://github.com/ruby/lrama/pull/536/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R45 because it's "parameterizing" when users define new parameterized rules.
Other seems good to me.

yui-knk (approved):

ydah:
> I agree to use "parameterized" in code base, actually "parameterized" is used in menhir code base. One point need consideration is the name of this feature https://github.com/ruby/lrama/pull/536/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R45 because it's "parameterizing" when users define new parameterized rules. Other seems good to me.

@yui-knk Indeed it is. We have reverted the function name to remain “Parameterizing”.

yui-knk:
LGTM

lib/lrama/warnings/redefined_rules.rb
```diff
@@ -14,8 +14,8 @@ def initialize(logger, warnings)
       def warn(grammar)
         return unless @warnings
 
-        grammar.parameterizing_rule_resolver.redefined_rules.each do |rule|
-          @logger.warn("parameterizing rule redefined: #{rule}")
+        grammar.parameterized_resolver.redefined_rules.each do |rule|
+          @logger.warn("parameterized rule redefined: #{rule}")
```
ydah, line 18: MEMO) I had a little difficulty here. Since this is a warning against “parameterized rules” rather than function names, I did not use Parameterizing. I don't have any strong opinions on this, so I will revert back if it feels different.

PR #574, opened 2025-03-01
Fix ielr tests

junk0612:


lib/lrama/states.rb
```diff
@@ -672,6 +669,7 @@ def split_states
       end
 
       compute_inadequacy_annotations
+      # @states.each {|state| state.item_lookahead_set.tap { p _1.transform_keys(&:to_s) }.clear }
```
yui-knk, line 672: Is it need to keep this commented out code?

lib/lrama/states.rb
```diff
@@ -631,17 +630,15 @@ def compute_always_follows
...
+        related_gotos = nterm_transitions.select {|goto2| has_internal_relation?(goto1, goto2) || has_successor_relation?(goto1, goto2) }
         [goto1, related_gotos]
       }.to_h
     end
 
     # @rbs () -> Hash[State::Action::Goto, bitmap]
     def compute_transition_bitmaps
       nterm_transitions.map {|goto|
-        [goto, Bitmap.from_array(goto.to_state.transitions.map {|transition| transition.next_sym.number })]
+        [goto, Bitmap.from_array(goto.to_state.term_transitions.map {|transition| transition.next_sym.number })]
```
yui-knk, line 641: [nits] It's better to use `shift` as block parameter instead of `transition`.

yui-knk (approved):

junk0612:
🙏 thanks for your review!

Issue #568, opened 2025-02-27
Support Start-Symbol: `%start`


ydah:
https://www.gnu.org/software/bison/manual/html_node/Start-Decl.html

PR #576, opened 2025-03-01
Add support Start-Symbol: `%start`

ydah:
Fixes: https://github.com/ruby/lrama/issues/568

### DIFF
--- a/lib/lrama/grammar.rb
+++ b/lib/lrama/grammar.rb
@@ -66,7 +66,7 @@ module Lrama
     #   @required: bool
     #   @union: Union
     #   @precedences: Array[Precedence]
-    #   @start_nterms: Array[Lrama::Lexer::Token]
+    #   @start_nterm: Lrama::Lexer::Token?
 
     extend Forwardable
 
@@ -131,7 +131,7 @@ module Lrama
       @define = define
       @required = false
       @precedences = []
-      @start_nterms = []
+      @start_nterm = nil
 
       append_special_symbols
     end
@@ -186,9 +186,14 @@ module Lrama
       set_precedence(sym, Precedence.new(s_value: s_value, type: :precedence, precedence: precedence, lineno: lineno))
     end
 
-    # @rbs (Lrama::Lexer::Token id) -> Array[Lrama::Lexer::Token]
-    def add_start_nterms(id)
-      @start_nterms << id
+    # @rbs (Lrama::Lexer::Token id) -> Lrama::Lexer::Token
+    def set_start_nterm(id)
+      if @start_nterm.nil?
+        @start_nterm = id
+      else
+        start = @start_nterm #: Lrama::Lexer::Token
+        raise "Start non-terminal is already set to #{start.s_value} (line: #{start.first_line}). Cannot set to #{id.s_value} (line: #{id.first_line})."
+      end
     end
 
     # @rbs (Grammar::Symbol sym, Precedence precedence) -> (Precedence | bot)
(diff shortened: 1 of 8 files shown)


yui-knk:
I tried the grammar file below with Bison 3.8.2.

```yacc
%{
#include <stdio.h>
#include <stdlib.h>
#include <ctype.h>
%}

%code provides {
static int yylex(YYSTYPE *val, YYLTYPE *loc);
static int yyerror(YYLTYPE *loc, const char *str);
}

%union {
    int val;
}
%token LF
%token <val> NUM
%type <val> expr
%left '+' '-'
%left '*' '/'

%locations

%start program_1 program_2 program_3 '/'

%%

program_1 : expr '+' expr;
program_2 : expr '-' expr;
program_3 : expr '*' expr;


expr : NUM
     | '(' expr ')'  { $$ = $2; }
     ;

%%
```

Output file says `YY_PARSE_` terminals are required on state 0. Could you check how multiple start symbols are used?

```
State 0

    0 $accept: • YY_PARSE_program_1 program_1 $end
    1        | • YY_PARSE_program_2 program_2 $end
    2        | • YY_PARSE_program_3 program_3 $end
    3        | • YY_PARSE_(NULL) '/' $end

    YY_PARSE_(NULL)     shift, and go to state 1
    YY_PARSE_program_1  shift, and go to state 2
    YY_PARSE_program_2  shift, and go to state 3
    YY_PARSE_program_3  shift, and go to state 4
```

yui-knk:
JFYI: https://www.gnu.org/software/bison/manual/html_node/Multiple-start_002dsymbols.html

ydah:
>Output file says YY_PARSE_ terminals are required on state 0. Could you check how multiple start symbols are used?

This is because Bison did not generate an error, so I made the same behavior, but if it is not supported, it is better to make it an error, so I changed it.

ydah:
@yui-knk An error is now raised if more than one `%start` is defined. Please point out if there is something wrong with your understanding of the sentence.

lib/lrama/grammar.rb
```diff
@@ -446,6 +452,19 @@ def normalize_rules
       @rules.sort_by!(&:id).freeze
     end
 
+    # Add $accept rule to the top of rules
+    def add_accept_rules
```
yui-knk, line 456: This method adds single accept rule then it's better to use `add_accept_rule` (without "s") or something like that.

yui-knk (approved):
Left comments. The direction looks good to me.

lib/lrama/grammar.rb
ydah: Change in https://github.com/ruby/lrama/pull/576/commits/4337cce0087f6d4d1562bbc94343ddf7f2927210

lib/lrama/grammar.rb
```diff
@@ -184,6 +186,16 @@ def add_precedence(sym, precedence, s_value, lineno)
       set_precedence(sym, Precedence.new(s_value: s_value, type: :precedence, precedence: precedence, lineno: lineno))
     end
 
+    # @rbs (Lrama::Lexer::Token id) -> Lrama::Lexer::Token
+    def set_start_nterm(id)
+      if @start_nterm.nil?
+        @start_nterm = id
+      else
```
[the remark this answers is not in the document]
ydah: I added the following comments. What do you think?
- https://github.com/ruby/lrama/pull/576/commits/0869f333342501b99a23ac6ce84e4daa6850655b

yui-knk (approved):

PR #582, opened 2025-03-01
Refactor Lrama::Reporter::States

ydah:


lib/lrama/reporter/states.rb
```diff
@@ -15,273 +15,337 @@ def initialize(itemsets: false, lookaheads: false, solved: false, counterexample
 
       # @rbs (IO io, Lrama::States states) -> void
       def report(io, states)
-        if @counterexamples
-          cex = Counterexamples.new(states)
-        end
+        cex = Counterexamples.new(states) if @counterexamples
```
yui-knk, line 18: We can move this instance creation into `#report_counterexamples`.

yui-knk: Oh, it's wrong. `#report_counterexamples` is called for each `state` then `cex` is needed to allocate outside of the loop.

lib/lrama/reporter/states.rb
```diff
@@ -15,273 +15,337 @@ def initialize(itemsets: false, lookaheads: false, solved: false, counterexample
...
 
         states.states.each do |state|
           # Report State
```
yui-knk, line 21: This comment is for `io << "State #{state.id}\n\n"` then can remove now.

yui-knk (approved):

lib/lrama/reporter/states.rb
ydah: Thank you! remove in: https://github.com/ruby/lrama/pull/582/commits/bfa580159e8e51fe70aa5cf5c50c24f9ba9975c6

PR #586, opened 2025-03-02
Fix tracer class type declarations

yui-knk:
This PR include two kind of fixes.

1. Fix type of rest keyword parameter `**Type` means type of keyword parameter value. Then `**Hash[Symbol, bool]` means `Symbol: Hash[Symbol, bool]`. What these methods expect is `Symbol: bool` so `**bool` is correct.

2. Declare optional keyword with `?key: bool` This correct way to declare optional keyword parameter.


ydah (approved):
Thank you!

PR #591, opened 2025-03-02
Enable `Lrama::ERB` type checking in Steepfile

ydah:


lib/lrama/erb.rb
```diff
@@ -5,7 +5,7 @@
 
 module Lrama
   class ERB
-    # @rbs (String file, Hash[Symbol, untyped]) -> string
+    # @rbs (String file, **Hash[Symbol, untyped] kwargs) -> String
```
yui-knk, line 8: I think `**untyped kwargs` is correct.

yui-knk: See 1 of https://github.com/ruby/lrama/pull/586

yui-knk (approved):

lib/lrama/erb.rb
ydah: oh... thank you :bow:

PR #593, opened 2025-03-03
Refactor "lib/lrama/reporter/states.rb"

yui-knk:
* Removing local variable type annotations by removing intermediate tuple local variables
* Remove needless `#uniq` method call
* Move early return to the first


ydah (approved):
Thank you!

PR #595, opened 2025-03-04
Optimize `compute_ielr`

junk0612:
Optimizing compute_ielr.
When I ran it in my local environment as a test, the execution time was reduced to about 6-7 minutes.

```
$ time bundle exec lrama -D lr.type=ielr parse.y
bundle exec lrama -D lr.type=ielr parse.y  377.91s user 0.67s system 99% cpu 6:18.84 total
```

### DIFF
--- a/lib/lrama/state.rb
+++ b/lib/lrama/state.rb
@@ -42,6 +42,7 @@ module Lrama
     attr_reader :items #: Array[States::Item]
     attr_reader :annotation_list #: Array[InadequacyAnnotation]
     attr_reader :predecessors #: Array[State]
+    attr_reader :items_to_state #: Hash[Array[States::Item], State]
 
     attr_accessor :_transitions #: Array[[Grammar::Symbol, Array[States::Item]]]
     attr_accessor :reduces #: Array[Action::Reduce]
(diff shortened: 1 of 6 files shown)


lib/lrama/state.rb
```diff
@@ -330,12 +347,21 @@ def annotate_predecessor(predecessor)
             [action, cs]
           end
         }.to_h
-        if (at = predecessor.annotation_list.find {|a| a.state == annotation.state && a.token == annotation.token && a.actions == annotation.actions })
-          at.merge_matrix(contribution_matrix)
-        else
-          predecessor.annotation_list << InadequacyAnnotation.new(annotation.state, annotation.token, annotation.actions, contribution_matrix)
-        end
+
+        new_annotation = InadequacyAnnotation.new(annotation.state, annotation.token, annotation.actions, contribution_matrix)
```
yui-knk, line 351: No need to assign to a local variable.

lib/lrama/states/item.rb
```diff
@@ -114,7 +114,7 @@ def display_rest
 
       # @rbs (States::Item other_item) -> bool
       def predecessor_item_of?(other_item)
-        rule == other_item.rule && position == other_item.position - 1
+        position == other_item.position - 1 && rule == other_item.rule
```
yui-knk, line 117: Just clarification, is there any major performance difference between checking `position` first and `rule` first?

yui-knk (approved):

lib/lrama/states/item.rb
junk0612: Oops. This was a trial and error. It didn't seem to affect the speed when I put it back.

PR #634, opened 2025-03-18
Fix behavior for R/R conflicts

junk0612:


lib/lrama/states.rb
```diff
@@ -791,5 +792,13 @@ def rr_conflicts_within_threshold?(logger, expected: 0)
       logger.error("reduce/reduce conflicts: #{rr_conflicts_count} found, #{expected} expected")
       false
     end
+
+    # @rbs () -> void
+    def clear_look_ahead_sets
```
yui-knk, line 797: How do you think about clear all instance variables used to LA set computation?

```
@direct_read_sets
@reads_relation
@read_sets
@includes_relation
@lookback_relation
@follow_sets
@la
```

junk0612: I thought that clearing the necessary cache would be sufficient.
I recognize that @direct_read_sets, @read_sets and @follow_sets will update as they are re-computed.

yui-knk: It's correct however these values can't be used in the computation for IELR so it's safer to clear the value.

junk0612: That's true. fixed.

yui-knk (approved):

PR #639, opened 2025-03-18
Fix RBS type annotations for Node class

ydah:


lib/lrama/counterexamples/node.rb
```diff
@@ -8,7 +8,7 @@ class Node
       attr_reader :elem #: E
       attr_reader :next_node #: Node[E]?
 
-      # @rbs　[T] (Node[T] node) -> Array[T]
```
ydah, line 11: memo) There was a full-width space in it, so perhaps the rbs parsing was failing.

yui-knk (approved):

PR #641, opened 2025-03-18
Update transition caches instead of clearing

junk0612:
Clearing the entire cache will adversely affect subsequent fetches of follow_kernel_items, etc.

### DIFF
--- a/lib/lrama/state.rb
+++ b/lib/lrama/state.rb
@@ -158,14 +158,25 @@ module Lrama
     def update_transition(transition, next_state)
       set_items_to_state(transition.to_items, next_state)
       next_state.append_predecessor(self)
-      clear_transitions_cache
+      update_transitions_caches(transition)
     end
 
     # @rbs () -> void
-    def clear_transitions_cache
+    def update_transitions_caches(transition)
+      new_transition =
+        if transition.next_sym.term?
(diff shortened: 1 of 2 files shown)


lib/lrama/state.rb
```diff
@@ -158,14 +158,26 @@ def transitions
...
+
+      @transitions.delete(transition)
+      @transitions << new_transition
       @nterm_transitions = nil
       @term_transitions = nil
-      @transitions = nil
+
+      @follow_kernel_items[new_transition] = @follow_kernel_items.delete(transition)
+      @always_follows[new_transition] = @always_follows.delete(transition)
+      @goto_follows[new_transition] = @goto_follows.delete(transition)
```
yui-knk, line 180: I'm wondering it's correct to inherit `goto_follows` of the original transition because `goto_follows` could change by state split.

junk0612: I think that's true. Because goto_follow_set is only called from inside item_lookahead_set and should not be called once it enters the split_state phase.

yui-knk: If `@goto_follows[new_transition]` is never called, I recommend to delete this line to make `@goto_follows[new_transition]` to return `nil` so that we can notice something unexpected happens as soon as possible.

yui-knk (approved):

Issue #480, opened 2024-11-12
Create parser without defining %union


ydah:
If %union is not defined, Integer is used by default.

```c
union YYSTYPE
{
#line 20 "sample/parse.y"

    int i;

#line 129 "parse.tab.c"

};
typedef union YYSTYPE YYSTYPE;
```

PR #654, opened 2025-03-18
Create parser without defining `%union`

ydah:
Fixes: https://github.com/ruby/lrama/issues/480

If %union is not defined, Integer is used by default.

```c
/* Value type.  */
#if ! defined YYSTYPE && ! defined YYSTYPE_IS_DECLARED
typedef int YYSTYPE;
# define YYSTYPE_IS_TRIVIAL 1
# define YYSTYPE_IS_DECLARED 1
#endif
```


yui-knk:
Do we need to take care of "template/bison/_yacc.h" if `union` is nullable?

ydah:
>Do we need to take care of "template/bison/_yacc.h" if union is nullable?

I tried to take care of it. How about this?

parser.y
```diff
@@ -6,7 +6,7 @@ class Lrama::Parser
 
 rule
 
-  input: prologue_declaration* bison_declaration* "%%" rules_or_grammar_declaration+ epilogue_declaration?
+  input: prologue_declaration* bison_declaration* ("%%" rules_or_grammar_declaration+ epilogue_declaration?)?
```
yui-knk, line 9: I think the first `%%` is mandatory.

ydah: Agreed, I updated this PR.

parser.y
```diff
@@ -6,7 +6,7 @@ class Lrama::Parser
 
 rule
 
-  input: prologue_declaration* bison_declaration* "%%" rules_or_grammar_declaration+ epilogue_declaration?
+  input: prologue_declaration* bison_declaration* "%%" (rules_or_grammar_declaration+ epilogue_declaration?)?
```
yui-knk, line 9: Could you keep at least one `rules_or_grammar_declaration` is required then do not change the grammar on this PR?
"lib/lrama/grammar/stdlib.y" can be valid if `%%` on line 142 is moved to line 10 or around because `%rule ...` can be both `bison_declaration` and `rules_or_grammar_declaration` then they are interpreted as `rules_or_grammar_declaration`.

ydah: ok. I revert it.

yui-knk:
Could you squash the commits into one?

yui-knk (approved):

PR #684, opened 2025-05-24
Fix MSYS2 setup by disabling auto-update and adding manual recovery steps

ydah:
Fixes: https://github.com/ruby/lrama/actions/runs/15225718911/job/42827501889?pr=683


yui-knk (approved):
Thank you!

Issue #674, opened 2025-04-13
An error occurs if`%left <i> FOO <j> BAR`


ydah:
https://github.com/Perl/perl5/blob/2adc4025ac962d6debca427c98bde824bc4dc4d7/perly.y#L135C1-L135C53

```yacc
%left <ival> OROR DORDOR <pval> PLUGIN_LOGICAL_OR_OP
%left <ival> ANDAND <pval> PLUGIN_LOGICAL_AND_OP
```

PR #688, opened 2025-06-14
Add support for allowing multiple tags to be mixed on one line in precedence declarations

ydah:
Fixes: https://github.com/ruby/lrama/issues/674


yui-knk (approved):
Please squash commits before merge this PR.

PR #693, opened 2025-07-27
Temporarily uses ruby 3.4 to avoid "uninitialized constant RBS::Collection::Sources::Git::FileUtils (NameError)"

yui-knk:
Ref: https://github.com/ruby/lrama/actions/runs/16546542959/job/46795014272


yui-knk:
Note: We can revert this change once https://github.com/ruby/rbs/commit/c9389681acf10c9fdd34601e74153011919b8544 is released and we update rbs version.

PR #704, opened 2025-08-01
Fix confusing error messages when specifying non-terminators for `%precedence`

ydah:
Before:

```
❯ exe/lrama sample/json.y
[BUG] LHS of object -> '{' option_members '}' (line: 38) is term. It should be nterm.
```

After:

```
❯ exe/lrama sample/json.y
[BUG] Precedence of object (line: 28) is not term. It should be term.
```

### DIFF
--- a/lib/lrama/grammar.rb
+++ b/lib/lrama/grammar.rb
@@ -161,24 +161,24 @@ module Lrama
       @types << Type.new(id: id, tag: tag)
     end
 
-    # @rbs (Grammar::Symbol sym, Integer precedence) -> Precedence
-    def add_nonassoc(sym, precedence)
-      set_precedence(sym, Precedence.new(type: :nonassoc, precedence: precedence))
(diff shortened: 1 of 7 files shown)


lib/lrama/grammar.rb
```diff
@@ -507,6 +508,21 @@ def fill_sym_to_rules
       end
     end
 
+    # @rbs () -> void
+    def validate_precedence_is_term!
```
yui-knk, line 512: I think it might be better to name this method as  `#validate_no_precedence_for_nterm`.

lib/lrama/grammar.rb
```diff
@@ -507,6 +508,21 @@ def fill_sym_to_rules
...
+      errors = [] #: Array[String]
+
+      @rules.each do |rule|
```
yui-knk, line 515: I feel it makes the intention more clear if we iterate over the `@nterms` even so all of non terminals should appear on the some rule's LHS.

ydah:
Thanks for the review. I will create a follow-up PR.

PR #711, opened 2025-08-02
Add support for precedence handling and warnings for unused precedences

ydah:
Example1:
```
%{
#include <stdio.h>
#include <stdlib.h>

int yylex(void);
void yyerror(const char *s);
%}

%union {
    int i;
}

%token NUMBER
%precedence PRECEDENCE

%%

program:
    NUMBER %prec PRECEDENCE
    ;

%%

int main(void) {
    printf("Parser ready. Enter expressions:\n");
    return yyparse();
}

void yyerror(const char *s) {
    fprintf(stderr, "Error: %s\n", s);
}

```

Before:
```
❯ exe/lrama -W tmp/example.y

```

After:
```
❯ exe/lrama -W tmp/example.y
Precedence PRECEDENCE (line: 14) is defined but not used in any rule.
```

Example2:
```
%{
#include <stdio.h>
#include <stdlib.h>

int yylex(void);
void yyerror(const char *s);
%}

%union {
    int i;
}

%token NUMBER
%precedence PRECEDENCE

%%

program:
    NUMBER
    ;

%%

int main(void) {
    printf("Parser ready. Enter expressions:\n");
    return yyparse();
}

void yyerror(const char *s) {
    fprintf(stderr, "Error: %s\n", s);
}

```

Before:
```
❯ exe/lrama -W tmp/example.y

```

After:
```
❯ exe/lrama -W tmp/example.y
Precedence PRECEDENCE (line: 14) is defined but not used in any rule.
```

### DIFF
--- a/lib/lrama/warnings.rb
+++ b/lib/lrama/warnings.rb
@@ -4,6 +4,7 @@
 require_relative 'warnings/conflicts'
 require_relative 'warnings/redefined_rules'
 require_relative 'warnings/required'
+require_relative 'warnings/useless_precedence'
 
 module Lrama
   class Warnings
@@ -12,6 +13,7 @@ module Lrama
       @conflicts = Conflicts.new(logger, warnings)
       @redefined_rules = RedefinedRules.new(logger, warnings)
       @required = Required.new(logger, warnings)
+      @useless_precedence = UselessPrecedence.new(logger, warnings)
     end
 
     # @rbs (Lrama::Grammar grammar, Lrama::States states) -> void
@@ -19,6 +21,7 @@ module Lrama
       @conflicts.warn(states)
       @redefined_rules.warn(grammar)
       @required.warn(grammar)
+      @useless_precedence.warn(grammar, states)
     end
   end
 end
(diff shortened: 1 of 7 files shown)


yui-knk:
I agree with the idea of issuing a warning for unused precedence.
However, I believe the current implementation for determining if a precedence is unused is incorrect.
Precedence is used to resolve conflicts.
Therefore, whether a precedence is "used" should be determined by whether it's actually resolving a conflict.
In Lrama, conflict resolution happens on [these codes](https://github.com/ruby/lrama/blob/11deaed2910756b9da30b891af341b13cce7b82a/lib/lrama/states.rb#L503-L545).
Then I suggest to

1. Add a new filed to `Precedence` such as `used` to track the precedence is used to resolve some conflicts in `compute_shift_reduce_conflicts`
2. Check the `used` flag in `UselessPrecedence` class
3. Update the test case to include
   1. precedence which is used to resolve conflict
   2. precedence which is not used to resolve conflict then warned
   3. precedence which is used with `%prec` declaration on some rules then not warned

PR #714, opened 2025-08-04
Update NEWS for v0.7.1

yui-knk:


NEWS.md
```diff
@@ -2,6 +2,55 @@
...
+### Add support for reporting Rule Usage Frequency
+
+https://github.com/ruby/lrama/pull/677
+
+### Render Split States information on output file
+
+https://github.com/ruby/lrama/pull/624
+
+### Add ioption support to the Standard library
+
```
ydah, line 51:
```suggestion
Support `ioption` (inline option) rule, which is expanded inline without creating intermediate rules.

Unlike the regular `option` rule that generates a separate rule, `ioption` directly expands at the point of use:

```yacc
program: ioption(number) expr

// Expanded inline to:

program: expr
       | number expr
```

This differs from the regular `option` which would generate:

```yacc
program: option(number) expr

// Expanded to:

program: option_number expr
option_number: %empty
             | number
```

The `ioption` rule provides more compact grammar generation by avoiding intermediate rule creation, which can be beneficial for reducing the parser's rule count and potentially improving performance.

This feature is inspired by Menhir's standard library and maintains compatibility with Menhir's `ioption` behavior.

```

PR #718, opened 2025-08-11
Change to raise an error if `%precedence` is set more than once for the same terminal symbol

ydah:
Example:
```
%{
// Prologue
%}
%union {
    int i;
}
%left tSTRING tNUMBER
%precedence tSTRING
%right tIDENTIFIER
%nonassoc tSTRING tNUMBER
%%
program: tNUMBER
        ;
```

Before:
```
$ exe/lrama example.y

```

After:
```
❯ exe/lrama tmp/a.y
[BUG] %precedence redeclaration for tSTRING (line: 8) previous declaration was %left (line: 7)
[BUG] %nonassoc redeclaration for tSTRING (line: 10) previous declaration was %left (line: 7)
[BUG] %nonassoc redeclaration for tNUMBER (line: 10) previous declaration was %left (line: 7)
```

### DIFF
--- a/lib/lrama/grammar.rb
+++ b/lib/lrama/grammar.rb
@@ -65,6 +65,7 @@ module Lrama
     #   @define: Hash[String, String]
     #   @required: bool
     #   @union: Union
+    #   @precedences: Array[Precedence]
 
     extend Forwardable
 
@@ -75,6 +76,7 @@ module Lrama
     attr_reader :accept_symbol #: Grammar::Symbol
     attr_reader :aux #: Auxiliary
     attr_reader :parameterized_resolver #: Parameterized::Resolver
+    attr_reader :precedences #: Array[Precedence]
     attr_accessor :union #: Union
     attr_accessor :expect #: Integer
     attr_accessor :printers #: Array[Printer]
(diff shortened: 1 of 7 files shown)


lib/lrama/grammar.rb
```diff
@@ -538,6 +543,25 @@ def validate_rule_lhs_is_nterm!
...
 
+    # # @rbs () -> void
+    def validate_duplicated_precedence!
+      errors = [] #: Array[String]
+      seen = {} #: Hash[String, Precedence]
+
+      precedences.each do |prec|
+        s_value = prec.s_value
+        if first = seen[s_value]
+          errors << "[BUG] %#{prec.type} redeclaration for #{s_value} (line: #{prec.lineno}) previous declaration was %#{first.type} (line: #{first.lineno})"
```
yui-knk, line 554: Could you remove "[BUG]" from the error message?
I use "[BUG]" to indicate the situation which Lrama would not expect, e.g. precedence is specified for nonterminal symbol. However redeclaration of precedence can happen by the user side.

ydah: I see. I updated error message.

PR #732, opened 2025-08-24
Rename `midrule_action` to `action` in parser rules for clarity

ydah:
A midrule-action is an action in the middle of a rule, and this nonterminal symbol is actually includes the trailing action. In practice, action is a better name. WDYT?


yui-knk (approved):
Sound reasonable.

PR #733, opened 2025-08-24
Fix an action in last rule of token definition for Precedence

ydah:
Example
```
%left ADDOP <pval> PLUGIN_ADD_OP
```

Before:
```
parser.y:328:in 'Lrama::Parser#_reduce_99': undefined method 'append' for an instance of Lrama::Lexer::Token::Ident (NoMethodError)
```

After:
```
// No error
```


parser.y
```diff
@@ -325,7 +325,7 @@ rule
   token_declarations_for_precedence:
       id+ { result = [{tag: nil, tokens: val[0]}] }
     | (TAG id+)+ { result = val[0].map {|tag, ids| {tag: tag, tokens: ids} } }
-    | id TAG id+ { result = val[0].append({tag: val[1], tokens: val[2]}) }
+    | id TAG id+ { result = [{tag: nil, tokens: [val[0]]}, {tag: val[1], tokens: val[2]}] }
```
yui-knk, line 328: I guess this rule might be `id+ TAG id+` then the id before `TAG` can be a list.
Could you double check it?

ydah: Thank you for catching this! I fixed it.

yui-knk (approved):

PR #748, opened 2025-09-08
Fix line number counting when there are trailing spaces

mame:
Previously, when there are trailing whitespace characters at the end of line, lrama miscounted the line number.

```
  // <=== Place whitespace here (and do not place this comment)
%{puts("Hello");
%}
%%
a: ‘-’;
```

```
$ ./exe/lrama t.y && less y.tab.c
*snip*

/* First part of user prologue.  */
#line 1 "t.y"  // <=== This should be #line 2
puts("Hello");

*snip*
```


ydah (approved):
Thank you!

